PR #45, opened 2024-08-21
feat: support creation of WlRegion, used for setting input region

nairaner:
I wanted to create overlay "transparent"  for mouse/input events, but wasn't able to create WlRegion.

### DIFF
--- a/layershellev/examples/simplelayer.rs
+++ b/layershellev/examples/simplelayer.rs
@@ -34,6 +34,17 @@ fn main() {
                         .unwrap(),
                 );
                 println!("{:?}", virtual_keyboard_manager);
+                ReturnData::RequestCompositor
+            },
+            LayerEvent::CompositorProvide(compositor, qh) => {
+                // NOTE: you can set input region to limit area which gets input events
+                // surface outside region becomes transparent for input events
+                // To ignore all input events use region with (0,0) size
+                for x in ev.get_unit_iter() {
+                    let region = compositor.create_region(&qh, ());
+                    region.add(0, 0, 0, 0);
+                    x.get_wlsurface().set_input_region(Some(&region));
+                }
                 ReturnData::None
             }
             LayerEvent::XdgInfoChanged(_) => {
--- a/layershellev/src/events.rs
+++ b/layershellev/src/events.rs
@@ -7,6 +7,7 @@ use wayland_client::{
     globals::GlobalList,
     protocol::{
         wl_buffer::WlBuffer,
+        wl_compositor::WlCompositor,
         wl_output::WlOutput,
         wl_pointer::{self, ButtonState, WlPointer},
         wl_shm::WlShm,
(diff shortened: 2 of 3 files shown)


layershellev/examples/simplelayer.rs
```diff
@@ -23,6 +23,15 @@ fn main() {
             // NOTE: this will send when init, you can request bind extra object from here
             LayerEvent::InitRequest => ReturnData::RequestBind,
             LayerEvent::BindProvide(globals, qh) => {
+                // Example: pass through input events, useful for overlays
+                // use ::wayland_client::protocol::wl_compositor;
+                // if let Some(compositor) = globals.bind::<wl_compositor::WlCompositor, _, _>(&qh, 1..=5, ()).ok() {
+                //     if let Some(state) = ev.get_unit_with_id(index.unwrap()) {
```
Decodetalkers, line 29: Wl_compositer is already in the `running`, why not add another returnvalue for it?

layershellev/src/lib.rs
```diff
@@ -1387,6 +1388,18 @@ impl<T> Dispatch<xdg_surface::XdgSurface, ()> for WindowState<T> {
         }
     }
 }
+
+impl<T> Dispatch<WlRegion, ()> for WindowState<T> {
```
Decodetalkers, line 1392: I think it can use dispatch_noop! macro here

Decodetalkers:
Thank you very much!

layershellev/src/lib.rs
nairaner: Thanks, I didn't understand how that worked, it seems better

layershellev/src/events.rs
```diff
@@ -121,6 +126,7 @@ pub enum ReturnData<INFO> {
     WlBuffer(WlBuffer),
     RequestBind,
     RequestExist,
```
nairaner, line 128: I think this should be called `RequestExit` ?

nairaner:
Thanks for feedback. I didn't understand code well enough to figure out how to connect all components. Now it should be better.

layershellev/src/events.rs
Decodetalkers: Yes, it is a typo, you can help fix it for me

layershellev/src/lib.rs
```diff
@@ -1824,6 +1826,13 @@ impl<T: 'static> WindowState<T> {
                         None,
                     ));
                 }
+                Some(ReturnData::RequestCompositor) => {
```
Decodetalkers, line 1829: Emm. I mean add another return value like `RequestSetRegion`... and it should not be in this place I think

Decodetalkers:
Emm. must the region be set only once time? if just once, I think this pr is ok

Decodetalkers:
ok. I will merge it now. thank you for your effort

PR #57, opened 2024-09-18
feat: prepaire for iced_13

Decodetalkers:


Decodetalkers:
NOTE: this one is a very beginning support for iced.. the code is really shit with lots of part which I cannot understand. but finall it run..I use tryinto to uniquely handle the message, but make it clone to be clone able.Maybe we should restrict the error of tryinto to be the message itself, but I do not know how to do

PR #65, opened 2024-09-26
added support of enum visibility to the `to_layer_message` attribute

tukanoidd:
Hi, updated the library in a personal project of mine, and due to how I structured my project, supporting my AppMsg visibility is important, so I added the feature in :)

### DIFF
--- a/iced_layershell_macros/src/lib.rs
+++ b/iced_layershell_macros/src/lib.rs
@@ -37,6 +37,7 @@ pub fn to_layer_message(attr: TokenStream, input: TokenStream) -> TokenStream {
         }
     }
     // Extract the enum name
+    let enum_vis = &input_enum.vis;
     let enum_name = &input_enum.ident;
     let variants = &input_enum.variants;
     let new_varents = if is_multi {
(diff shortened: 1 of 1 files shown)


Decodetalkers:
I see, this is the mark of pub pub(crate), right?

Decodetalkers (approved):

tukanoidd:
> I see, this is the mark of pub pub(crate), right?

yep

Decodetalkers:
And do you know a better way to pass derive to the derive made enum? now I set the derive as input.. but I do not think it is the best way...

Decodetalkers:
Thank you very much!!

tukanoidd:
> And do you know a better way to pass derive to the derive made enum? now I set the derive as input.. but I do not think it is the best way...

Ye, I prefer to use this crate for typed parsing of macros https://docs.rs/darling/latest/darling/ and it works well with syn, quote and proc-macro2

tukanoidd:
I could help with a refactor when i have free time :)

Decodetalkers:
Thank you!

Issue #71, opened 2024-10-03
Blurry interface using layershell


MalpenZibo:
Text on the interface seems blurry compared to the one obtained using "standard" iced or the pop_os fork (to access Wayland layershell).

Could my HiDPI monitor cause it (Resolution: 2880x1800)? 
I'm using Hyprland with a 2.0 scaling)

![screenshot-20241003-10:54:07](https://github.com/user-attachments/assets/6a8fb0ba-c0f0-42d8-9e90-b193d4d969e9)

Decodetalkers:
I have not handle the case for scale.. I do not know how to handle it will be better

Decodetalkers:
I have made a pr https://github.com/waycrate/exwlshelleventloop/pull/73 , but there is something I cannot solved.. the size of viewport.. I do not quite familiar with iced_graphic.. I need some help

MalpenZibo:
I will try my best to help you!

MalpenZibo:
The blurry effect is gone but now the application renders double the size. There should be somewhere an option to say to the iced layout system to limit the size of the main container.

Decodetalkers:
> The blurry effect is gone but now the application renders double the size. There should be somewhere an option to say to the iced layout system to limit the size of the main container.

This is what I do not know.. I do not know much about iced.. I try do what iced_winit do, but the size of viewport is wrong,, I do not know what happened, so I need help

MalpenZibo:
I'm looking here `https://github.com/MalpenZibo/iced_sctk` this is an """hard fork""" of the work done by the pop_os team on their fork of iced. I'm currently using this code to run my status bar. I have just removed the pop_os customization on iced maintaining only the support for the layershell protocol to use iced 12. 

I notice that they set up the surface with the physical size (for example in my configuration a width of 1440 px creates a surface with a width of 2880 px) and when they create the Viewport they use a physical size and a scale (in my case 2880 px and 2.0 of scaling) but then they call
```rust
self.viewport = Viewport::with_physical_size(physical_size, scale);
if let Some(wp_viewport) = self.wp_viewport.as_ref() {
  wp_viewport.set_destination(
    logical_size.width.ceil() as i32,
    logical_size.height.ceil() as i32,
  );
}
```

I'm not an expert but the wp_viewport is a `WpViewport` that seems a Wayland protocol. This is what I found reading the Rust wayland-protocol library documentation
```
pub mod viewporter {
    //! Viewporter protocol
    //!
[...]

PR #73, opened 2024-10-04
feat:  support fractal scale

Decodetalkers:
resolve: https://github.com/waycrate/exwlshelleventloop/issues/71


Decodetalkers:
The size of window is little strange

PR #97, opened 2024-12-02
feat: add build_pattern for iced_layershell

Decodetalkers:


Decodetalkers:
try to add build pattern similar to iced for iced_layershell

PR #109, opened 2024-12-21
Fix event transparency not working

Exidex:
I might have not tested properly when I initially requested this feature. Here's a fix to get it actually working


Decodetalkers:
Thanks

Issue #110, opened 2024-12-22
Set input_region for surface


nebneb0703:
I'm using iced_layershell for my application. I have overridden the default style to use a transparent background but my app contains inner opaque elements. The transparent regions of my application need to be transparent to input events, corresponding to the input_region of the wl_surface. The actual gui is dynamic and therefore I cannot rely on a static layershell size. I could not find a suitable api accessible through iced_layershell.

I have two suggestions for solutions:

1. Provide a new action capable of performing this change at runtime. The input region is part of the core wayland protocol, not the layershell protocol, so it might not entirely make sense to include it alongside other parts of this crate.
2. Provide an escape hatch by exposing the wl_surface of the current window, such that clients can perform the set_input_region request manually.

Although not necessary for this issue, it might be useful to consider support for setting the opaque_region as well.

Decodetalkers:
Yesterday this was fixed by a contributor, so maybe you can test the master branch, if it is good, I will make a release

nebneb0703:
If you are referring to #109, it does not do what I need. The events_transparent flag is a bool that applies to the whole surface, and there is no way to inject a specific region. It's also not dynamic and only applies once upon creation of the surface.

nebneb0703:
Added in #111, released in v0.13.1

PR #111, opened 2024-12-27
Added basic support for input_regions

manthanabc:
# Attempts to solve #110 
This creates an interface to dynamically set input region, which can be edited by user by means of a new action
```rust
   // Simple application
   LayershellCustomActoin::SetInputRegion(|region| {
            region.add(0, 0, 5000, 5000);
            region.subtract(0, 0, 50, 50);
    }))
   // Multi-Window application
   Message::SetInputRegion { 
      set_region: |region| { 
        region.add(500,500,1920,1080); 
      }, id: WindowId
   }
```
I'm marking this as draft, any suggestions would be much appreciate since its my first time contributing here.
This also adds nessasory changes to to_layer_message macro, and to multi window mode.

### DIFF
--- a/iced_layershell/README.md
+++ b/iced_layershell/README.md
@@ -468,4 +468,83 @@ impl Counter {
 
 ```
 
+### Dynamically set input region
+```rust, no_run
+
+use iced::widget::{button, row, Space};
+use iced::{Color, Element, Length, Task as Command, Theme};
+use iced_layershell::settings::{LayerShellSettings, Settings};
+use iced_layershell::to_layer_message;
+use iced_layershell::Application;
+
+pub fn main() -> Result<(), iced_layershell::Error> {
+    InputRegionExample::run(Settings {
+        layer_settings: LayerShellSettings {
+            size: Some((400, 400)),
+            ..Default::default()
+        },
+        ..Default::default()
+    })
+}
+
+struct InputRegionExample;
+
+#[to_layer_message]
+#[derive(Debug, Clone)]
+#[doc = "Some docs"]
+enum Message {
+    SetRegion,
+    UnsetRegion,
+}
+
+impl Application for InputRegionExample {
+    type Message = Message;
+    type Flags = ();
+    type Theme = Theme;
+    type Executor = iced::executor::Default;
+
+    fn new(_flags: ()) -> (Self, Command<Message>) {
+        (Self, Command::none())
+    }
+
+    fn namespace(&self) -> String {
+        String::from("Counter - Iced")
+    }
+
+    fn update(&mut self, message: Message) -> Command<Message> {
+        match message {
+            Message::SetRegion => Command::done(Message::SetInputRegion(|region| {
+                // Only the buttons!
+                region.add(0, 0, 400, 70);
+            })),
+            Message::UnsetRegion => Command::done(Message::SetInputRegion(|region| {
+                // Entire window!
+                region.add(0, 0, 400, 400);
+            })),
+            _ => unreachable!(),
+        }
+    }
+
+    fn view(&self) -> Element<Message> {
+        // Create the top row with two buttons
+        row![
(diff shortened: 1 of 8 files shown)


iced_layershell/src/multi_window.rs
```diff
@@ -219,28 +221,33 @@ where
...
-                    file,
-                    keymap_size,
-                    keymap_format,
-                } = settings.virtual_keyboard_support.as_ref().unwrap();
-                let seat = ev.get_seat();
-                let virtual_keyboard_in =
-                    virtual_keyboard_manager.create_virtual_keyboard(seat, qh, ());
-                virtual_keyboard_in.keymap((*keymap_format).into(), file.as_fd(), *keymap_size);
-                ev.set_virtual_keyboard(virtual_keyboard_in);
+                let wl_compositor = globals.bind::<WlCompositor, _, _>(qh, 1..=1, ());
```
Decodetalkers, line 227: Here can use expect, because wl_compositor should exist

iced_layershell/src/application.rs
```diff
@@ -269,6 +275,22 @@ where
                     LayershellCustomActions::LayerChange(layer) => {
                         ev.main_window().set_layer(layer);
                     }
+                    LayershellCustomActions::SetInputRegion(set_region) => {
+                        let window = ev.main_window();
+
+                        if let Some(region) = &wl_input_region {
```
Decodetalkers, line 281: This will exist after doing bind, so here can also use expect

Decodetalkers:
Can you add an example for this new apo?

iced_layershell/README.md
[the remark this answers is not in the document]
Decodetalkers: I mean add an examples folder and add this example instead. And it will be better to add comments in the place of this code

Decodetalkers:
I think it is already fine to be merged

Decodetalkers:
Now finally , you need to fix the clippy problem

manthanabc:
> I think it is already fine to be merged

Yea, but we can't yet pass closures to set_input, unless it does not capture the environment variables (so are automatically converted to fn). which seems really useful, since without it the shapes have to be pretty much statically decided at compile time. to allow for it i would either need to 
- change SetInputRegion(&'a (dyn Fn(&WlRegion)) which would mean i'd need to introduce the <'a> everywhere LayershellCustomAction is contained which i dont really like
- or just wrrape it in box<dyn Fn(&WlRegion)> which would need user to also wrap the closure in box which is fine but it does heap allocation so might not be desirable
i cannot decide which is i should do

manthanabc:
@Decodetalkers   does this look good :eyes: ??

iced_layershell/src/actions.rs
```diff
@@ -94,7 +113,6 @@ impl LayershellCustomActionsWithId {
 }
 
 // first one means
-#[derive(Debug, Clone, Copy)]
```
Decodetalkers, line 97: There is no need to delete this derive, I think just deleting the copy is enough

iced_layershell/src/actions.rs
```diff
@@ -82,9 +104,6 @@ pub enum LayershellCustomActions {
     ForgetLastOutput,
 }
 
-/// Please do not use this struct directly
```
Decodetalkers, line 85: this comment should not be deleted

manthanabc: yeah, i must have removed it accidentally.

iced_layershell/src/actions.rs
manthanabc: i forgot to add it back

iced_layershell/rust-analyzer
Decodetalkers, line 1: what is this?

manthanabc: i have no clue how it got there, fixing it

Decodetalkers:
ok, I will merge it

PR #118, opened 2025-01-17
fix: scaling issue.

fortime:
1. set destination of wpviewport in the creation of a state.
2. all width and height from wayland should be logical.
3. move logic of resizing viewport into `resize_viewport`.

This PR is currently applied to iced_layershell/multi_window. If it is ok, i will work on other parts.

### DIFF
--- a/iced_layershell/src/multi_window/state.rs
+++ b/iced_layershell/src/multi_window/state.rs
@@ -16,7 +16,6 @@ where
     id: window::Id,
     application_scale_factor: f64,
     wayland_scale_factor: f64,
-    real_window_size: Size<u32>,
     viewport: Viewport,
     viewport_version: usize,
     theme: A::Theme,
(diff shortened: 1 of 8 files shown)


iced_layershell/src/multi_window/state.rs
```diff
@@ -41,62 +40,38 @@ where
...
 
-        let real_window_size = Size::new(width, height);
-        let viewport = Viewport::with_physical_size(
-            real_window_size,
-            wayland_scale_factor * application_scale_factor,
-        );
+        let logical_size = Size::new(width, height);
+        let viewport = viewport(logical_size, wayland_scale_factor, application_scale_factor);
+
+        let wpviewport = window
```
Decodetalkers, line 46: At the beginning the size maybe 0,0 so I wait for the next time for the size being adjusted and set the destination

Decodetalkers:
Seems good, let me take a try

Decodetalkers:
Thank you for your work!

iced_layershell/src/multi_window/state.rs
fortime: The size should be provided by application. In which case will it be (0, 0)? Also, if the size has never been changed, the viewport will be wrong.

Decodetalkers: for the iced_layershell to make a compositor, it need a waylandWrapper, at this time I cannot know the size, but when the surface is configured, it will update. so my design here is wait for the later update to update the viewport

fortime: Sorry. I don't quite understand this case. I'm new to Wayland. I can't find `waylandWrapper` in this repo. Will this new code cause any issue in this situation?

Decodetalkers: Emm.. what you have done is better.. but I think here , you should better to check if the size is (0,0 )

For example:

https://github.com/waycrate/exwlshelleventloop/blob/master/iced_layershell/src/multi_window.rs#L193

The window here is not configured, so the size is 0, 0, so this window need wait the next time that surface is configured to update the viewport

Decodetalkers:
And maybe you need to run the format here

iced_layershell/src/multi_window/state.rs
fortime: Oh, I see. When I work with `daemon` to create a background process, it creates a fullscreen window even I don't send a `NewLayerShell` message. I think at this case the size will be (0, 0).

BTW, I want to create another PR. I think the default `StartMode` for daemon should be `Background`.

Back to this part, you mean I should check the size before `set_destination`? What happens if the application sends a `NewLayerShell` message with size (0, 0)?

Decodetalkers: Background is just for background programs..and there are many programs. Because it does not use any wl_surface binded with the GUI, but is an extra one,  which used to create the composer, to avoid when gui is closed the composer also died. So this results in me needing to keep updating the surface , because it will never be received with an Outdate stated.  For a background program it may be enough, because it won't show for a long time.. but for a normal program, it consults a lot of resources.

Decodetalkers:
buy the way, can you also fix that in iced_sessionlock? thanks

fortime:
> buy the way, can you also fix that in iced_sessionlock? thanks

Done. But I'm using kwin, it doesn't support sessionlock. I'm not sure whether it is correct.

Decodetalkers:
ok, I will test tomorrow. seems good for me!

Decodetalkers:
good for me! I merge it!

PR #131, opened 2025-02-27
fix: surface reconfigured every frame

ksdrar:
Cache surface sizes and scale to avoid configuring the surface every frame, this reduces CPU usage.

### DIFF
--- a/iced_layershell/src/multi_window.rs
+++ b/iced_layershell/src/multi_window.rs
@@ -524,7 +524,8 @@ async fn run_instance<A, E, C>(
         };
     }
     let mut window_manager: WindowManager<A, C> = WindowManager::new();
-    let mut sizes: HashMap<iced_core::window::Id, iced_core::Size<u32>> = HashMap::new();
+    let mut cached_layer_dimensions: HashMap<iced_core::window::Id, (iced_core::Size<u32>, f64)> =
+        HashMap::new();
 
     let mut clipboard = LayerShellClipboard::unconnected();
     let mut ui_caches: HashMap<window::Id, user_interface::Cache> = HashMap::new();
@@ -685,13 +686,14 @@ async fn run_instance<A, E, C>(
 
                 let physical_size = window.state.physical_size();
 
-                if match sizes.get(&id) {
+                if match cached_layer_dimensions.get(&id) {
                     None => true,
-                    Some(size) => {
-                        size.height != physical_size.height || size.width != physical_size.width
+                    Some((size, scale)) => {
+                        *size != physical_size || *scale != window.state.scale_factor()
                     }
                 } {
-                    sizes.insert(id, physical_size);
+                    cached_layer_dimensions
+                        .insert(id, (physical_size, window.state.scale_factor()));
 
                     compositor.configure_surface(
                         &mut window.surface,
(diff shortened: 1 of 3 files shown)


ksdrar:
Testing got my app from 6% CPU usage down to 0.5%

iced_layershell/src/multi_window.rs
```diff
@@ -524,6 +524,7 @@ async fn run_instance<A, E, C>(
         };
     }
     let mut window_manager: WindowManager<A, C> = WindowManager::new();
+    let mut sizes: HashMap<layershellev::id::Id, iced_core::Size<u32>> = HashMap::new();
```
Decodetalkers, line 527: I prefer to use the id o iced

Decodetalkers:
can you make the same change in iced_sessionlock?

iced_layershell/src/multi_window.rs
```diff
@@ -683,11 +684,21 @@ async fn run_instance<A, E, C>(
...
-                compositor.configure_surface(
-                    &mut window.surface,
-                    physical_size.width,
-                    physical_size.height,
-                );
+
+                if match sizes.get(&window.id) {
+                    None => true,
+                    Some(size) => {
+                        size.height != physical_size.height || size.width != physical_size.width
```
Decodetalkers, line 691: I think here should also store the scale

Decodetalkers: size can be directly compared I think

iced_layershell/src/multi_window.rs
```diff
@@ -524,6 +524,7 @@ async fn run_instance<A, E, C>(
         };
     }
     let mut window_manager: WindowManager<A, C> = WindowManager::new();
+    let mut sizes: HashMap<iced_core::window::Id, iced_core::Size<u32>> = HashMap::new();
```
Decodetalkers, line 527: I think here should create a new struct, to store the information of size and scale.

iced_layershell/src/multi_window.rs
Decodetalkers: I think here should also check if it has need to relayout, and this part just check the configure surface

iced_layershell/src/multi_window.rs
```diff
@@ -683,11 +685,22 @@ async fn run_instance<A, E, C>(
                 }
 
                 let physical_size = window.state.physical_size();
-                compositor.configure_surface(
-                    &mut window.surface,
-                    physical_size.width,
-                    physical_size.height,
-                );
+
+                if match cached_layer_dimensions.get(&id) {
```
Decodetalkers, line 689: here can be replaced with is_some_and

Decodetalkers: should also check the relayout above I think

iced_layershell/src/multi_window.rs
ksdrar: I think a tuple is enough for this small use case

Decodetalkers:
And there is similar logic in iced_sessionlock, can you also fix it there?

ksdrar:
Sure! Working on it

ksdrar:
I think that covers all the requests

Decodetalkers (approved):

Decodetalkers:
Thank you very much!

PR #133, opened 2025-03-05
feat: add support for changing the exclusive zone size.

fortime:
Add an variant `ExclusiveZoneChange` to change the exclusive zone size.

### DIFF
--- a/iced_layershell/src/multi_window.rs
+++ b/iced_layershell/src/multi_window.rs
@@ -364,6 +364,7 @@ where
                         }
                         LayershellCustomActions::ExclusiveZoneChange(zone_size) => {
                             let Some(id) = id else {
+                                tracing::error!("Here should be an id, it is a bug, please report an issue for us");
                                 break 'out;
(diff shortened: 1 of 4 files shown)


iced_layershell/src/multi_window.rs
```diff
@@ -362,6 +362,15 @@ where
                             };
                             window.set_size((width, height));
                         }
+                        LayershellCustomActions::ExclusiveZoneChange(zone_size) => {
+                            let Some(id) = id else {
```
Decodetalkers, line 366: I think here should also log an error as the other place, because here the id should exist.

Decodetalkers (approved):

iced_layershell/src/multi_window.rs
fortime: Done. This code is written in Jan. So it didn't contain the log like other branches.

Decodetalkers:
Thank you very much

PR #136, opened 2025-03-07
feat(multi_window): send `iced::window::Event::Closed`.

fortime:
Send `iced::window::Event::Closed` after a window is removed. This will be more consistent with the behavior of `iced_winit`. It also make handling after-closed tasks more easily compared to `remove_id`.

### DIFF
--- a/iced_layershell/src/multi_window.rs
+++ b/iced_layershell/src/multi_window.rs
@@ -919,6 +919,11 @@ async fn run_instance<A, E, C>(
                     &mut window_manager,
                     cached_user_interfaces,
                 ));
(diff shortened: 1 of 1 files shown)


Decodetalkers (approved):

Decodetalkers:
thank you very much

Decodetalkers:
I found that you are writing a virtualkeyboard program, so should I publish a new version for you? That is really a good project

fortime:
Thank you. But currently there is no need for a new version. My project is still WIP. It won't be released in the near future.
BTW, there are two PR on the way😆. And I am debugging a panic during closing the first window of multi_window.

PR #138, opened 2025-03-09
Iced14

Decodetalkers:


Decodetalkers:
problem is , preedit will never show, and text_input won't work normally in multi_window

Decodetalkers:
I will not support the debug recently, keep it this time will be just ok..when I have time I will implement the debug feature

Decodetalkers:
because of new boot design and debug design , the old trait cannot be kept anymore, so we need to remove it

Decodetalkers:
I believe that iced14 won't make any breaking change anymore, it is safe to make it merged

PR #144, opened 2025-03-11
fix: there is no effect on changing the application scale factor.

fortime:
While I was copying the `multi_window` example from `iced` to this repo, I found that there is no effect on changing the application scale factor.

In this commit, a `window_size` is introduced in the window state. There are three  types of coordinates:
    1. `window_size`: The coordinates that the client believes it is using. These are used to communicate with the Wayland server.
    2. `viewport.physical_size`: The coordinates of the screen. These are used to render the final view that the user sees.
    3. `viewport.logical_size`: The coordinates used by Iced. Iced draws widgets based on these coordinates.

```
viewport.physical_size = window_size * wayland_scale_factor = viewport.logical_size * application_scale_factor * wayland_scale_factor
 ```

I also removed methods `physical_size` and `logical_size` from window state, because they are a bit confused.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2421,6 +2421,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "multi_window"
+version = "0.13.5"
+dependencies = [
+ "iced",
+ "iced_layershell",
+ "iced_runtime",
+ "tracing",
+ "tracing-subscriber",
+]
+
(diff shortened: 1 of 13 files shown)


Decodetalkers (approved):

Decodetalkers:
Thank you very much

PR #149, opened 2025-03-25
Feat/multitouch

fortime:
1. fix: inconsistent surface id between TouchDown event and TouchUp event in multi window.
2. fix: incorrect cursor position in TouchUp event.
3. feat: support multi touch.
4. feat: Dont't reset current_surface when TouchUp event and MouseLeave event happened. So keyboard events can still be sent to the last active surface instead of the first surface in the vector.

### DIFF
--- a/sessionlockev/src/lib.rs
+++ b/sessionlockev/src/lib.rs
@@ -125,7 +125,7 @@ use wayland_client::{
         wl_display::WlDisplay,
         wl_keyboard::{self, KeyState, KeymapFormat, WlKeyboard},
         wl_output::{self, WlOutput},
-        wl_pointer::{self, WlPointer},
+        wl_pointer::{self, ButtonState, WlPointer},
         wl_registry,
         wl_seat::{self, WlSeat},
         wl_shm::WlShm,
@@ -159,6 +159,7 @@ use wayland_protocols::wp::fractional_scale::v1::client::{
     wp_fractional_scale_v1::{self, WpFractionalScaleV1},
 };
 
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::sync::Mutex;
 use std::sync::atomic::AtomicBool;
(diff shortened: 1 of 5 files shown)


sessionlockev/src/lib.rs
```diff
@@ -1048,15 +1119,6 @@ impl<T> Dispatch<wl_pointer::WlPointer, ()> for WindowState<T> {
                     },
                 ));
             }
-            wl_pointer::Event::Leave { .. } => {
```
Decodetalkers, line 1051: why do we not need leave anymore? This part I am little concerned about..

sessionlockev/src/lib.rs
```diff
@@ -1048,15 +1119,6 @@ impl<T> Dispatch<wl_pointer::WlPointer, ()> for WindowState<T> {
...
-                state.current_surface = None;
-                state
-                    .message
-                    .push((state.surface_id(), DispatchMessageInner::MouseLeave));
-                if let Some(keyboard_state) = state.keyboard_state.as_mut() {
-                    keyboard_state.current_repeat = None;
```
Decodetalkers, line 1057: if this part is not been produced.. I am afraid that the keyboard repeating will be of bugs..

Decodetalkers: I see, you move it above.. sorry for that

Decodetalkers:
seems all good, but now master is freezed for iced14.. I think iced14 will been published soon, so can you try to merged it to iced14 branch? Thanks

fortime:
> sorry, I want to you to pr to the iced14 branch, I hope the master branch be freezed until iced14 is published..

OK. I will try creating a new PR to iced14. It may not be that soon.

fortime:
There are some bugs in this pr, the logic in handling current surface can be better. And I found the cursor logic during RequestRefresh has some problems in multi windows. During current implementation, each window has its cursor during refreshing. This is not true.

I'm closing this pr and working on a new fix. After that, I will port it to iced14 branch and create a new pr.

PR #152, opened 2025-04-15
fixs of touch event within multi windows

fortime:
This is a revision of #149 . In kwin6, I found touch events can be happened in different surfaces at the same time. So instead of updating `current_surface`, I store active surfaces in a map.

1. inconsistent surface of touch event.
2. incorrect position of touch event.
3. update `current_surface` only if a finger is down or a mouse button is clicked or a surface is created.
4. update cursor shape only if the surface is hovered by the mouse.
5. put the logic of setting cursor shape into a function.

### DIFF
--- a/iced_examples/multi_window/src/main.rs
+++ b/iced_examples/multi_window/src/main.rs
@@ -78,7 +78,7 @@ impl Example {
             7 => Anchor::Left | Anchor::Bottom,
             _ => Anchor::Bottom,
         };
-        let size = (1024, 768);
+        let size = (480, 320);
         let id = window::Id::unique();
         (
             id,
--- a/iced_layershell/src/application.rs
+++ b/iced_layershell/src/application.rs
@@ -203,10 +203,8 @@ where
 
     let mut context = task::Context::from_waker(task::noop_waker_ref());
     let mut wl_input_region: Option<WlRegion> = None;
-    let mut pointer_serial: u32 = 0;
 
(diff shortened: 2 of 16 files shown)


Decodetalkers:
seems good for me! thank you for your effort!

Decodetalkers:
seems iced made break change again... can you help me modify the code? it changed a lot. these time I do not have many time to follow..

fortime:
> seems iced made break change again... can you help me modify the code? it changed a lot. these time I do not have many time to follow..

I don't quite understand what you mean. Do you mean that after updating the version of iced the code can't be compiled?
Is it related to this PR? If not, maybe we can discuss this in other place, such as opening a discussion.

Decodetalkers:
> > seems iced made break change again... can you help me modify the code? it changed a lot. these time I do not have many time to follow..
> 
> I don't quite understand what you mean. Do you mean that after updating the version of iced the code can't be compiled? Is it related to this PR? If not, maybe we can discuss this in other place, such as opening a discussion.

Emm.. just try to search some help.. but still I can fix it myself.. but I just want some help..

fortime:
I may help, but I don't know how. So, I suggest that  you should open an issue or discussion to discuss it. Also, with a new topic, others can participate in this.
BTW, I may help, but I am not sure when it will be finished. I am going to work out how my virtual keyboard works in SDDM.

PR #154, opened 2025-04-19
Move iced features to separate feature flags

WhySoBad:
Hi,

This pull request disables the default features of the `iced` crate and moves them to separate feature flags for the `iced_layershell` and `iced_sessionlock` crates. 

The reason for this is simple: As soon as the `wgpu` feature flag is enabled, wgpu is used as the primary renderer and tiny-skia can only be used as a fallback renderer. One can use tiny-skia as primary renderer only when the `wgpu` feature flag is disabled and the `tiny-skia` feature flag is enabled. 

On my (hyprland) system, I noticed that wgpu does no damage tracking and re-renders the whole layer surface on every frame which results in a performance penalty. Additionally, tiny-skia has a much lower memory footprint which is also preferrable. 

With the changes of this pull request it's possible for users of this crate to disable the default features and only turn on the `tiny-skia` feature. This isn't possible with the current implementation. I've moved all default features of `iced` into the default features of this crate and therefore the default behavior should stay the same

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,7 +67,7 @@ rwh_05 = { package = "raw-window-handle", version = "0.5.2", features = [
 ] }
 rwh_06 = { package = "raw-window-handle", version = "0.6", features = ["std"] }
 
-iced = { git = "https://github.com/iced-rs/iced.git" }
+iced = { git = "https://github.com/iced-rs/iced.git", features = ["auto-detect-theme"], default-features = false }
 iced_runtime = { git = "https://github.com/iced-rs/iced.git" }
 #iced_style = "0.13"
 iced_core = { git = "https://github.com/iced-rs/iced.git" }
--- a/iced_layershell/Cargo.toml
+++ b/iced_layershell/Cargo.toml
@@ -10,6 +10,12 @@ readme = "README.md"
 description = "layershell binding for iced"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[features]
+default = ["tiny-skia", "wgpu", "fira-sans"]
+tiny-skia = ["iced/tiny-skia", "iced_renderer/tiny-skia"]
+wgpu = ["iced/wgpu", "iced_renderer/wgpu"]
+fira-sans = ["iced/fira-sans", "iced_renderer/fira-sans"]
+
 [dependencies]
 iced.workspace = true
 iced_renderer.workspace = true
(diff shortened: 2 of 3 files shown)


Decodetalkers:
I think you would better enable it in your project, but not in the lib, I do not think it should be set the flags here.

WhySoBad:
Well yeah, I'd agree with you. 

However, I have set up my project to use tiny-skia (no appearance of `wgpu` in `cargo tree`) and as soon as I add the `iced_layershell` dependency `wgpu` appears in my cargo tree which then results in my application running on wgpu.

Decodetalkers:
> Well yeah, I'd agree with you. 
> 
> However, I have set up my project to use tiny-skia (no appearance of `wgpu` in `cargo tree`) and as soon as I add the `iced_layershell` dependency `wgpu` appears in my cargo tree which then results in my application running on wgpu.

I see, then can you make pr to iced14 branch? I am considering merging that branch after iced14 is published, so I would like to froze the master branch

WhySoBad:
Done. 

One thing I noticed is that you changed the `iced` dependencies to use git urls. As you probably want to update the `iced_layershell` and `iced_sessionlock` crates on crates.io, keep in mind that using git dependencies is not allowed on crates.io: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-other-registries

Decodetalkers:
> Done. 
> 
> One thing I noticed is that you changed the `iced` dependencies to use git urls. As you probably want to update the `iced_layershell` and `iced_sessionlock` crates on crates.io, keep in mind that using git dependencies is not allowed on crates.io: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-other-registries

After iced14 is published, I will remove the git://

Decodetalkers:
But I do not think the crate included in the lib will influence your project.. it is weird. I think if you disable the feature in the iced of your project, it should have been disabled..

Decodetalkers (approved):

WhySoBad:
It's documented here: https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#feature-unification

Cargo unifies all features of the same crate, should the crate appear multiple times in the dependency tree. Therefore, the features are additive which means the default features of iced take overhand.

Decodetalkers:
> It's documented here: https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#feature-unification
> 
> Cargo unifies all features of the same crate, should the crate appear multiple times in the dependency tree. Therefore, the features are additive which means the default features of iced take overhand. 

Thank you

PR #173, opened 2025-05-13
feat(layershell): support zwp_input_panel_v1

fortime:
1. support zwp_input_panel_v1.
2. add `with_connection` in `Settings` so that an application can extend their wayland implementation.

## Background of Adding `with_connection`
In kwin, only the connection opened from `WAYLAND_SOCKET` can use `zwp_input_panel_v1` and `zwp_input_method_v1`. So if we want to show a input panel and input something through this input panel, we must implement `zwp_input_panel_v1` and `zwp_input_method_v1` in the same program. With `with_connection`, an application can create a connection and register its wayland event queue.

### DIFF
--- a/iced_layershell/src/actions.rs
+++ b/iced_layershell/src/actions.rs
@@ -4,8 +4,8 @@ use enumflags2::BitFlags;
 use iced::window::Id as IcedId;
 use iced_core::input_method;
 use iced_core::mouse::Interaction;
-use layershellev::NewLayerShellSettings;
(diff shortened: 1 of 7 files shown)


Decodetalkers (approved):

Decodetalkers:
Thanks for your contribution

Decodetalkers:
Emm.. now it is the master branch.. can you make another pr for that? Sorry

Decodetalkers:
ok, let me cherry-pick it

PR #181, opened 2025-05-16
Move theme detection behind feature flag

WhySoBad:
Hi,

In #154 I intentionally left the `auto-detect-theme` feature flag as a iced default. However, I just realized that this depends on zbus and therefore on an async runtime. Since their default async runtime is `async-std` this will add a new, additional runtime for most projects. 

Therefore, I think it's better to be able to turn this off as it may cause a lot of hardware threads to be allocated for a feature which isn't needed (as it's the case for me)


Decodetalkers:
Thanks

PR #220, opened 2025-07-16
Commit input region changes

### LINKED ISSUE #219: Input region changes only apply after next message
Hi,

I'm trying to set a custom input region for my layer. However, I've noticed that the input region (at least for iced) only updates after the next message is handled. When running with `WAYLAND_DEBUG=1` I can see that the input region is correctly applied (and my `add` goes through), however I'm still able to click through the specified region until I trigger a new message to be sent:

Image

WhySoBad:
Instantaneously send a `wl_surface.commit` after a change to the input region of a surface has been done. 

closes #219


Decodetalkers (approved):

Decodetalkers:
Thank you

PR #288, opened 2025-10-19
Run `mundy` inside `Runtime` context like in `iced_winit`

Baelyk:
Just like in `iced_winit`: https://github.com/iced-rs/iced/commit/0b78f2aa5d0a741da79f3b2b0a91046c32587403.

Without this, mundy panics when using zbus with tokio because it lacks a runtime.


Decodetalkers:
Thank you!

PR #297, opened 2025-10-30
fix tty switch crash

### LINKED ISSUE #287: sessionlock panics on TTY change
When using the `simplelock` example or my own code under sway the application panics when switching TTYs and switching back to to the session. 

The error I am able to capture is:

```
wl_display#1: error 1: invalid method 1 (since 1 < 3), object wl_pointer#12
Io error: Invalid argument (os error 22)
Io error: Invalid argument (os error 22)
```

The code in question is here:

```rust
        event_loop
            .run(
                std::time::Duration::from_millis(20),
                &mut self,
                |_window_state| {
                    // Finally, this is where you can insert the processing you need
                    // to do do between each waiting event eg. drawing logic if
                    // you're doing a GUI app.
                },
            )
            .expect("Error during event loop!");
```

I've attempted to walk through the code and handle some `.unwrap()` calls, but have yet to locate anything obvious.

rustysec:
patched `sessionlockev` and `layershellev` to avoid crashing on tty switch. 

```rust
pointer.release()
```

needs a check for version >= 3.

fixes #287


Decodetalkers:
Thank you very much! I will take a try!

Decodetalkers:
But I think it maybe good. Ok , thank you for your help

rustysec:
... it works for me using `simplelock` and `simplelayer`. both crash without these changes and work as intended with the changes.

Issue #317, opened 2025-12-16
Unfocus Event gives wrong window Id when focusing to another window in the same iced program.


DJ-Laser:
When switching focus between two layer shell surfaces, the Unfocus event does not have the correct window Id. It seems like the Focus event is fired on the focused layer shell (or window), then the Unfocused event fires with the id of the newly focused surface, instead of the previous one. For some reason re-focusing the surface will then trigger the unfocus event, unless you first focus an ordinary window.
When focusing to an ordinary window not owned by the iced process, it works correctly.

Focusing another layer shell (the center layer shell closes on unfocus, the bottom bar just tracks events with no reaction):
https://github.com/user-attachments/assets/abc98c66-d8d2-484b-afcc-58297c46e97a

Focusing a window owned by the same iced process (same thing but the bar is now in an xdg window).
https://github.com/user-attachments/assets/2fbe333f-90b3-4423-858a-695e19665e4f

I have not yet tested this on other des/wms, however the fact that it works on othee xgd windows but not those in the same iced process makes me think it is an issue with iced_layershell

The code I used can be found at https://github.com/dj-laser/n16-shell/tree/debug (use the previous commit for the bar in a layer shell instead of an xdg window)

DJ-Laser:
I did some testing using hyprland, and it correctly unfocuses the window when switching to another surface, though it also immediately unfocuses the newly opened window if another surface owned by the iced program is focused, same as niri.

This leads me to think it might at least in part be a window manager quirk, though I am curious why niri doesn't properly unfocus only when switching to a window owned by the same program.

PR #318, opened 2025-12-16
chore: when keyboard enter the surface, we should also update surface

Decodetalkers:
resolve: https://github.com/waycrate/exwlshelleventloop/issues/317

We forgot the handle the keyboard enter event


Decodetalkers:
Though it will not solve the problem, but I still think this pr need to be merged, so

Issue #159, opened 2025-05-05
iced_layershell: continuously doing something despite nothing changing


WhySoBad:
Hi,

I'm experiencing a high CPU usage for my application using the `iced_layershell` crate of about 9% to 20%. I'm using the `iced14` branch and using the new `debug` feature flag I've attached `comet` to the application. I don't think it's an regression since I'm experiencing similar CPU usage on an old version of my application which uses `iced 0.13.1` but there I can't attach `comet`.

I've noticed that the application is constantly rendering (or doing something at least) since in `comet` the charts have constant new updates. When attaching `comet` to the `tour` example of the iced repository one can see that this application only does something if effectively something changes (e.g. cursor move, button press) and otherwise nothing happens.

**How to reproduce**:

iced tour example (as reference how it should be):

1. clone iced repository
2. run tour example
3. attach `comet` by pressing F12

iced layershell application launcher example (how it is with `iced_layershell`):

1. enable debug flag for `iced_layershell` dependency
2. run application launcher example
3. attach `comet` by pressing F12

I'd suspect that this constant action in `iced_layershell` is the cause for the high CPU usage

*I don't think this is the same as #78 as the other issue seems to be a problem in iced whilst this is a problem in this crate. However, the high CPU usage noticed in the other issue could originate from this problem*

Decodetalkers:
Can you try latest commit? I think I have fixed it

WhySoBad:
The issue persists independent of the feature flag being turned on or off

Decodetalkers:
have you follow https://github.com/waycrate/exwlshelleventloop/pull/138/commits/491704681a01a660278bf069cf026a35a3eed79d this commit?

WhySoBad:
yes, I'm on https://github.com/waycrate/exwlshelleventloop/commit/24c937e4c47ecc94fd46ef9cd86dbfe9c70a624f

Decodetalkers:
> yes, I'm on [24c937e](https://github.com/waycrate/exwlshelleventloop/commit/24c937e4c47ecc94fd46ef9cd86dbfe9c70a624f)

Then I cannot know what happened.. I do not have that energy now. I have follow the change in iced, just update when state changed

WhySoBad:
No problem, but I don't think it's an regression as the `master` branch of `iced_layershell` also has this high CPU usage.

Decodetalkers:
by the way, what is comet

WhySoBad:
[...]

PR #364, opened 2026-04-12
perf: use dynamic dispatch timeout instead of hardcoded 20ms

Magniquick:
## Summary

The event loop in `layershellev` uses a fixed 20ms timeout for `calloop::EventLoop::run()`, causing ~50 wakeups/sec even when fully idle. This replaces `event_loop.run()` with a manual `event_loop.dispatch()` loop that computes the timeout from each unit's `RefreshRequest` state:

- `NextFrame` → `Duration::ZERO` (wake immediately)
- `At(instant)` → remaining time until that instant
- `Wait` (all units) → `None` (sleep indefinitely until an external event arrives)

External events (Wayland fd, channels, calloop timers) still wake the loop regardless of timeout.

## Impact

Reduces idle CPU from ~10% to ~0.7% on Hyprland (measured with `perf stat` and `strace -c` over 10 seconds). The remaining ~0.7% comes from the `roundtrip()` sync traffic which is addressed in a separate PR.

Only `layershellev` is patched here — `exwlshellev` and `sessionlockev` have the same pattern and would benefit from the same change.

## Disclaimer

This was generated with the help of Claude Code. I've verified it reduces idle CPU on my machine and am currently using it, but there hasn't been a human code review beyond that. I have minimal Rust knowledge. Feel free to close this without looking at it if you'd rather approach the fix differently.

Relates to #159

### DIFF
--- a/layershellev/src/lib.rs
+++ b/layershellev/src/lib.rs
@@ -3508,10 +3508,10 @@ impl<T: 'static> WindowState<T> {
 
             false
         };
-        // Dynamic dispatch timeout: sleep indefinitely when no unit needs a
-        // refresh, wake instantly when a frame is pending.  External events
-        // (Wayland fd, channels, timers registered on the calloop handle)
-        // still wake the loop regardless of timeout.
+        // Dynamic dispatch timeout: compute the sleep duration from each
+        // unit's RefreshRequest rather than using a fixed interval.
+        // Based on the approach used by winit's Wayland event loop:
+        // https://github.com/rust-windowing/winit/blob/master/winit-wayland/src/event_loop/mod.rs#L242-L312
         loop {
             let timeout = state.raw.min_dispatch_timeout();
             event_loop.dispatch(timeout, &mut state)?;
--- a/layershellev/src/lib.rs
+++ b/layershellev/src/lib.rs
@@ -780,6 +780,16 @@ impl<T> WindowStateUnit<T> {
         }
     }
 
+    /// Returns the duration until this unit needs its next refresh,
+    /// or `None` if no refresh is pending (`Wait` state).
+    fn refresh_timeout(&self) -> Option<Duration> {
+        match self.request_flag.refresh {
+            RefreshRequest::NextFrame => Some(Duration::ZERO),
+            RefreshRequest::At(instant) => Some(instant.saturating_duration_since(Instant::now())),
+            RefreshRequest::Wait => None,
+        }
+    }
+
     pub fn take_present_slot(&mut self) -> bool {
         if !self.should_refresh() {
             return false;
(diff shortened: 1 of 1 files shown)


Magniquick:
As mentioned above really. Again, close if you dont wanna review the AI generated code. Decided to try a PR since afaik this should help everyone and I'd not need to vendor this manually.

Decodetalkers:
Seems ai get the code from https://github.com/rust-windowing/winit/blob/master/winit-wayland/src/event_loop/mod.rs#L242-L312

You should mention it in relative part

Although I do not that love ai generated code, I will also not close the pr. I will still read the pr, because you pay a lot in this feature, not only the meaning of efforts, but also money. 

But at least, I want to mark clearly, about where did ai get them

Magniquick:
Fair enough lol.
I dont like AI much either, I hate wasted cpu cycles more. 
also cheked lisencing real quick, shouldnt be an issue.
also about the similarity, it didnt clone or read the repo. prolly there in the training data tho.

Currently manually vendoring my version for https://github.com/Magniquick/bubcop, btw

Decodetalkers:
> The AI (Claude Code) did not clone or reference the winit source during this session — the approach was derived from profiling the binary with `perf`/`strace`, reading the layershellev and calloop APIs, and working out the fix from first principles. That said, the pattern is likely in Claude's training data from winit's codebase, so the similarity is unsurprising.
> 
> Running `cargo fmt` now and pushing.

so please mark them, claude use the code from winit, so please mark the origin source, in comment

Magniquick:
done. Lemme know if you want me to squash the commits btw.

Decodetalkers (approved):
Ok, thank you for your contribution. I also have read the code of winit carefully this time. Though it needs to be sync to other two crates, still, I think this pr can pass

Decodetalkers:
I do not that like to use copilot, so next time please do not call it. If you have called it, do not merge its suggestions, thanks. This time I will let you pass

Magniquick:
> I do not that like to use copilot, so next time please do not call it. If you have called it, do not merge its suggestions, thanks. This time I will let you pass

_Pretty_ sure that is something you set up and not me btw.

> Ok, thank you for your contribution. I also have read the code of winit carefully this time. Though it needs to be sync to other two crates, still, I think this pr can pass

Do you want prs for that or do you want me to do it ?

Decodetalkers:
> > I do not that like to use copilot, so next time please do not call it. If you have called it, do not merge its suggestions, thanks. This time I will let you pass
> 
> _Pretty_ sure that is something you set up and not me btw.
> 
> > Ok, thank you for your contribution. I also have read the code of winit carefully this time. Though it needs to be sync to other two crates, still, I think this pr can pass
> 
> Do you want prs for that or do you want me to do it ?

I will do that later

By the way, I did not set anything. So I forgot to block the copilot. It should be me that call the copilot, not you. Even I forgot to disable this feature

Thank you for your understanding. I really do not like it

Magniquick:
> By the way, I did not set anything. So I forgot to block the copilot. It should be me that call the copilot, not you. Even I forgot to disable this feature

I _don't_ like it either. but there is literally nothing I can do. I didn't call Copilot. PR authors can't trigger or disable it; it's a repo-level setting under Settings -> Code review ->  Copilot code reviews. GitHub may have enabled it by default. You should be able turn it off there.

https://docs.github.com/en/copilot/how-tos/use-copilot-agents/request-a-code-review/configure-automatic-review

and I didnt do @copilot (and cant):
image

Decodetalkers:
Ok. I see, I have already disable the options of copilot. Maybe it is pushed to us by github

PR #365, opened 2026-04-12
perf: replace idle roundtrip() with dispatch_pending() + flush()

Magniquick:
## Summary

Builds on #364. `event_queue_origin.roundtrip()` sends `wl_display.sync` on every loop iteration. The compositor responds with `wl_callback.done`, which wakes the calloop `WaylandSource`, creating a ~48 Hz feedback loop even when fully idle.

This replaces it with:
- `event_queue_origin.dispatch_pending()` — process already-buffered events without I/O
- `connection.flush()` — send any queued outgoing requests

The calloop `WaylandSource` already reads from the Wayland socket and routes events to all queues, so the explicit roundtrip synchronization is redundant for normal operation.

## Impact

Combined with #364, reduces idle CPU from ~0.7% to 0.0% (6 total syscalls over 10 seconds, measured with `strace -c` and `perf stat` on Hyprland).

## Risk

This changes the event processing model — events on `event_queue_origin` are now dispatched only when the `WaylandSource` has already read them, rather than being synchronously pulled via `roundtrip()`. This needs testing on compositors beyond Hyprland (Sway, cosmic-comp, etc.) to verify no edge cases break.

Only `layershellev` is patched here — `exwlshellev` and `sessionlockev` have the same pattern.

## Disclaimer

This was generated with the help of Claude Code. I've verified it reduces idle CPU on my machine and am currently using it, but there hasn't been a human code review beyond that. I have minimal Rust knowledge. Feel free to close this without looking at it if you'd rather approach the fix differently.

Relates to #159

### DIFF
--- a/layershellev/src/lib.rs
+++ b/layershellev/src/lib.rs
@@ -3518,13 +3518,21 @@ impl<T: 'static> WindowState<T> {
 
             let r_window_state = &mut state;
             let window_state = &mut r_window_state.raw;
-            // Process events already buffered by the calloop WaylandSource
-            // without sending wl_display.sync.  roundtrip() generates a
-            // sync→done pair every iteration (~48 Hz feedback loop even when
-            // idle).  dispatch_pending() + flush() is sufficient because the
-            // WaylandSource reads from the Wayland socket and routes events
-            // to all queues; we just need to dispatch what was buffered for
-            // event_queue_origin and flush any outgoing requests.
+            // Use dispatch_pending() + flush() instead of roundtrip().
+            //
+            // roundtrip() sends wl_display.sync every call, blocking until
+            // the compositor replies — creating a ~48 Hz feedback loop even
+            // when idle.  Its docs note it is meant for "initial setup" and
+            // one-off synchronisation points, not per-frame use:
+            // https://docs.rs/wayland-client/latest/wayland_client/struct.EventQueue.html#method.roundtrip
+            //
+            // dispatch_pending() processes events already buffered by the
+            // calloop WaylandSource (which reads the Wayland socket and
+            // routes events to all queues), without any I/O:
+            // https://docs.rs/wayland-client/latest/wayland_client/struct.EventQueue.html#method.dispatch_pending
+            //
+            // flush() sends any outgoing requests queued by event handlers:
+            // https://docs.rs/wayland-client/latest/wayland_client/struct.EventQueue.html#method.flush
             let _ = event_queue_origin.dispatch_pending(window_state);
             let _ = connection.flush();
             let event_handler = &mut r_window_state.fun;
--- a/layershellev/src/lib.rs
+++ b/layershellev/src/lib.rs
@@ -3518,7 +3518,15 @@ impl<T: 'static> WindowState<T> {
 
             let r_window_state = &mut state;
             let window_state = &mut r_window_state.raw;
-            let _ = event_queue_origin.roundtrip(window_state);
+            // Process events already buffered by the calloop WaylandSource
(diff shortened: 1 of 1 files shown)


Decodetalkers:
Why two same prs?

Magniquick:
> Why two same prs?

Not the same - #364 is the dynamic timeout (replaces hardcoded 20ms). This PR adds one more change on top: replacing roundtrip() with dispatch_pending() + flush() (the 8-line diff at line 3521). GitHub shows the combined diff because both target master. The actual delta  between them is just the roundtrip -> dispatch_pending swap. Split them because #364 is safe and this one might need broader compositor testing, according to claude.

Decodetalkers:
please rebase it

layershellev/src/lib.rs
```diff
@@ -3466,93 +3508,95 @@ impl<T: 'static> WindowState<T> {
...
+        loop {
+            let timeout = state.raw.min_dispatch_timeout();
+            event_loop.dispatch(timeout, &mut state)?;
+
+            let r_window_state = &mut state;
+            let window_state = &mut r_window_state.raw;
+            // Process events already buffered by the calloop WaylandSource
+            // without sending wl_display.sync.  roundtrip() generates a
+            // sync→done pair every iteration (~48 Hz feedback loop even when
+            // idle).  dispatch_pending() + flush() is sufficient because the
```
Decodetalkers: give me the source of document thanks

Magniquick: should be fine now.

Decodetalkers:
I will ask the authors of wayland-rs, I won't merge this pr for sometime. When I think it is ready, I will set it to available

layershellev/src/lib.rs
```diff
@@ -3518,7 +3518,22 @@ impl<T: 'static> WindowState<T> {
...
+            // https://docs.rs/wayland-client/latest/wayland_client/struct.EventQueue.html#method.roundtrip
+            //
+            // dispatch_pending() processes events already buffered by the
+            // calloop WaylandSource (which reads the Wayland socket and
+            // routes events to all queues), without any I/O:
+            // https://docs.rs/wayland-client/latest/wayland_client/struct.EventQueue.html#method.dispatch_pending
+            //
+            // flush() sends any outgoing requests queued by event handlers:
+            // https://docs.rs/wayland-client/latest/wayland_client/struct.EventQueue.html#method.flush
+            let _ = event_queue_origin.dispatch_pending(window_state);
```
Decodetalkers, line 3536: https://docs.rs/wayland-client/0.31.14/wayland_client/struct.EventQueue.html#method.dispatch_pending

why do not use queue.flush() here?

Decodetalkers:
And I am thinking why we need to flush. I read the code of slurp, it just do dispatch_pending, without flushing

I think we do not need flush, because we are not waiting to read something from the queue, just run the loop. If we want some callback, it will be better to use roundtrip. I have read the document this time, and read the code of slurp.

layershellev/src/lib.rs
Decodetalkers: Flush pending outgoing events to the server

This needs to be done regularly to ensure the server receives all your requests, though several dispatching methods do it implicitly (this is stated in their documentation when they do).

https://docs.rs/wayland-client/latest/wayland_client/struct.Connection.html#method.flush

seems meaningful. I see

Decodetalkers:
Ok, can confirm, I have read the code of slurp, and read the document of wayland and wayland-client, I think this is a right fix, at least it will never induce troubles, because everyone handle the events in this way.

It was my bad to use roundtrip all the place

Thank you for your contribution

PR #370, opened 2026-04-24
feat: add clipboard opt-out at init

obayemi:
Expose `disable_clipboard()` in iced_layershell, iced_exwlshell, and iced_sessionlock so apps with no text input can skip spawning the smithay-clipboard worker thread.

## Why ?

I'm working on seing how low I could get the CPU and RAM usage in a iced-layershell based status bar. I've got it to a point where the mere existence of the clipboard worker thread is a significant part of the idle cpu usage for the project. so I'm proposing a way to disable it for apps that do not require it.


Decodetalkers (approved):

Decodetalkers:
Reasonable for me, thanks

Issue #371, opened 2026-04-27
StartMode::Background is broken due to infinite dispatch timeout


ibrahimduran:
I recently made a switch to the master rev from the latest release and looks like the the following commit has caused a bug:
https://github.com/waycrate/exwlshelleventloop/commit/167a5709bb7388c12eec5e6a0b8727b31112d14e

To reproduce:
```sh
$ git checkout e93615cd3ee29420bde4eb46866b9563e3ec5e7b
$ cargo run --package=zbus_invoked_widget
$ busctl --user call zbus.iced.MyGreeter1 /org/zbus/MyGreeter1 org.zbus.MyGreeter1 SayHello s "foo"
# a new window appears correctly

$ git checkout master
$ cargo run --package=zbus_invoked_widget
$ busctl --user call zbus.iced.MyGreeter1 /org/zbus/MyGreeter1 org.zbus.MyGreeter1 SayHello s "foo"
# no window
```

I did a bit of debugging and looks like the infinite timeout for the dispatch call [here](https://github.com/waycrate/exwlshelleventloop/blob/167a5709bb7388c12eec5e6a0b8727b31112d14e/layershellev/src/lib.rs#L3505-L3506) causes program to get stuck due to lack of an initial window.

I'm yet to fully understand these stuff but I gave it a shot [here](https://github.com/waycrate/exwlshelleventloop/pull/372), not sure if it's a worthy solution.

Decodetalkers:
I am wondering when to publish a new release..

PR #372, opened 2026-04-27
fix: have a dispatch timeout fallback for the background mode

ibrahimduran:
#371

Added fallback dispatch timeout for the first tick to make Background start mode work.


Decodetalkers (approved):

Decodetalkers:
I see, thanks, it is reasonable! Thanks for quick fix!

PR #375, opened 2026-04-30
feat: add client side decoration control for XDG windows

httpsworldview:
This adds a client side decorations option to XDG window settings so applications can request client-side decorations when desired instead of always being forced to use server-side decor. Please let me know if this needs any changes, or if this isn't desired. Thanks for your consideration!


Decodetalkers (approved):
LGTM

Decodetalkers:
Thank you!

PR #385, opened 2026-06-06
feat: popups rework

id3v1669:
Why:
While I was trying to implement apptray, got into limitations with current state of lib. Client-side popup positioning was wrong near screen edges, couldn't nest or dismiss properly, and positioning relative to absolute cursor pos was pain without ugly workarounds.

Reference:
delegate placement to the compositor via `xdg_positioner` just as in gtk with easy dismissal via xdg_popup grab with `PopupDone`

Also old impl gave us only direction instead of anchor+gravity+flip. Exposing those also makes life easier.

Concern I have is that portions of `wayland_protocols` are exposed, but does that really matter since things like `WlOutput` are already exposed. 
Anyway I'm 99% sure I missed something, plz review and tell what needs to be changed. 

ps. testing that commit in puppet project went well

### DIFF
--- a/exwlshellev/src/events.rs
+++ b/exwlshellev/src/events.rs
@@ -3,6 +3,11 @@ use wayland_protocols_wlr::layer_shell::v1::client::{
     zwlr_layer_surface_v1::{Anchor, KeyboardInteractivity},
 };
 
+use wayland_protocols::xdg::shell::client::xdg_positioner::{
+    Anchor as PopupAnchor, ConstraintAdjustment as PopupConstraintAdjustment,
+    Gravity as PopupGravity,
+};
+
 use wayland_client::{
     QueueHandle, WEnum,
     globals::GlobalList,
@@ -106,10 +111,18 @@ pub struct NewLayerShellSettings {
 pub struct NewPopUpSettings {
     /// the size of the popup
     pub size: (u32, u32),
-    /// the position of the popup, relative to the he layersurface
-    pub position: (i32, i32),
-    /// It means where the popup is, on which surface. It is the id of that layershell
+    /// the id of the parent surface
     pub id: id::Id,
+    /// anchor rectangle in the parent surface's local coordinates (x, y, w, h)
+    pub anchor_rect: (i32, i32, i32, i32),
+    /// which point of the anchor rect the popup is anchored to
+    pub anchor: PopupAnchor,
+    /// the direction the popup grows from the anchor point
+    pub gravity: PopupGravity,
+    /// how the compositor may adjust (flip/slide/resize) the popup for off-screen cases
+    pub constraint_adjustment: PopupConstraintAdjustment,
+    /// Serial of the input event
+    pub grab_serial: Option<u32>,
 }
 /// Settings used to create a new xdg toplevel window.
 #[derive(Debug, PartialEq, Eq, Clone, Default)]
--- a/exwlshellev/src/events.rs
+++ b/exwlshellev/src/events.rs
@@ -106,6 +106,15 @@ pub struct NewLayerShellSettings {
     pub namespace: Option<String>,
 }
 
+/// How a popup is positioned relative to its parent surface.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum PopupPlacement {
+    /// anchor rectangle in the parent surface's local coordinates (x, y, w, h)
+    Anchored((i32, i32, i32, i32)),
+    /// Absolute position of the popup in the parent surface's local coordinates
+    Position((i32, i32)),
+}
+
 /// be used to create a new popup
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub struct NewPopUpSettings {
@@ -113,8 +122,8 @@ pub struct NewPopUpSettings {
     pub size: (u32, u32),
     /// the id of the parent surface
     pub id: id::Id,
-    /// anchor rectangle in the parent surface's local coordinates (x, y, w, h)
-    pub anchor_rect: (i32, i32, i32, i32),
+    /// How a popup is positioned relative to its parent surface.
+    pub placement: PopupPlacement,
     /// which point of the anchor rect the popup is anchored to
     pub anchor: PopupAnchor,
     /// the direction the popup grows from the anchor point
--- a/iced_examples/counter_multi/src/main.rs
+++ b/iced_examples/counter_multi/src/main.rs
@@ -3,13 +3,13 @@ use std::collections::HashMap;
 use iced::widget::{button, column, container, row, text, text_input};
 use iced::window::Id;
 use iced::{Alignment, Element, Event, Length, Task as Command, event};
-use iced_layershell::actions::{IcedNewMenuSettings, IcedXdgWindowSettings, MenuDirection};
+use iced_layershell::actions::{IcedNewPopupSettings, IcedXdgWindowSettings};
 use iced_runtime::window::Action as WindowAction;
 use iced_runtime::{Action, task};
 
 use iced_layershell::daemon;
 use iced_layershell::reexport::{
-    Anchor, KeyboardInteractivity, Layer, NewLayerShellSettings, OutputOption,
+    Anchor, KeyboardInteractivity, Layer, NewLayerShellSettings, OutputOption, PopupGravity,
 };
 use iced_layershell::settings::{LayerShellSettings, Settings, StartMode};
 use iced_layershell::to_layer_message;
@@ -168,15 +168,19 @@ impl Counter {
                         }
                     }
                     Event::Mouse(iced::mouse::Event::ButtonPressed(iced::mouse::Button::Right)) => {
-                        let id = iced::window::Id::unique();
-                        self.ids.insert(id, WindowInfo::PopUp);
-                        return Command::done(Message::NewMenu {
-                            settings: IcedNewMenuSettings {
-                                size: (100, 100),
-                                direction: MenuDirection::Up,
-                            },
-                            id,
-                        });
+                        if let Some(parent) = self.window_id(&WindowInfo::Left).copied() {
+                            let id = iced::window::Id::unique();
+                            self.ids.insert(id, WindowInfo::PopUp);
+                            return Command::done(Message::NewPopUp {
+                                settings: IcedNewPopupSettings::new(
+                                    parent,
+                                    (100, 100),
+                                    (0, 0, 1, 1),
+                                )
+                                .gravity(PopupGravity::TopRight),
+                                id,
+                            });
+                        }
                     }
                     _ => {}
                 }
--- a/iced_examples/counter_multi/src/main.rs
+++ b/iced_examples/counter_multi/src/main.rs
@@ -3,13 +3,13 @@ use std::collections::HashMap;
 use iced::widget::{button, column, container, row, text, text_input};
 use iced::window::Id;
 use iced::{Alignment, Element, Event, Length, Task as Command, event};
-use iced_layershell::actions::{IcedNewPopupSettings, IcedXdgWindowSettings};
+use iced_layershell::actions::{IcedNewMenuSettings, IcedXdgWindowSettings, MenuDirection};
 use iced_runtime::window::Action as WindowAction;
 use iced_runtime::{Action, task};
 
 use iced_layershell::daemon;
 use iced_layershell::reexport::{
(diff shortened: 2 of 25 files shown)


iced_layershell/src/build_pattern/daemon.md
```diff
@@ -138,15 +138,15 @@ impl Counter {
...
-                                size: (100, 100),
-                                direction: MenuDirection::Up,
-                            },
-                            id,
-                        });
+                        if let Some(parent) = self.window_id(&WindowInfo::Left).copied() {
+                            let id = iced::window::Id::unique();
+                            self.ids.insert(id, WindowInfo::PopUp);
+                            return Command::done(Message::NewPopUp {
+                                settings: IcedNewPopupSettings::new(parent, (100, 100), (0, 0, 1, 1))
```
Decodetalkers, line 145: This will be quite difficult to understand.. I am thinking about use a margin structure, and a size for this part, maybe we can use something like impl Into<Size>?

iced_wayland_subscriber/README.md
```diff
@@ -175,15 +175,15 @@ impl Counter {
...
-                        let id = iced::window::Id::unique();
-                        self.ids.insert(id, WindowInfo::PopUp);
-                        return Command::done(Message::NewMenu {
-                            settings: IcedNewMenuSettings {
-                                size: (100, 100),
-                                direction: MenuDirection::Up,
-                            },
-                            id,
-                        });
+                        if let Some(parent) = self.window_id(&WindowInfo::Left).copied() {
```
Decodetalkers, line 178: Does this mean popup will only be shown on Left Window?

iced_exwlshell/src/multi_window.rs
```diff
@@ -836,17 +836,29 @@ where
...
-                let Some(parent_layer_shell_id) = ev.current_surface_id() else {
+                let IcedNewPopupSettings {
+                    size,
+                    parent,
+                    anchor_rect,
+                    anchor,
+                    gravity,
+                    constraint_adjustment,
+                } = settings;
+                let Some(parent_layer_id) = self.window_manager.get(parent).map(|w| w.id) else {
```
Decodetalkers: I want the popup can just be shown above the active layer, I mean, when we are using popup, we do not need to care about what the layer is. so I still prefer that we can still use current_id for it. for example, if not set which layer to show, we need to use the current layer

exwlshellev/src/events.rs
```diff
@@ -106,10 +111,18 @@ pub struct NewLayerShellSettings {
 pub struct NewPopUpSettings {
     /// the size of the popup
     pub size: (u32, u32),
-    /// the position of the popup, relative to the he layersurface
-    pub position: (i32, i32),
-    /// It means where the popup is, on which surface. It is the id of that layershell
+    /// the id of the parent surface
     pub id: id::Id,
+    /// anchor rectangle in the parent surface's local coordinates (x, y, w, h)
```
Decodetalkers: I think it should be left, right, top, bottom?

iced_examples/counter_multi/src/main.rs
```diff
@@ -168,15 +168,19 @@ impl Counter {
...
-                        let id = iced::window::Id::unique();
-                        self.ids.insert(id, WindowInfo::PopUp);
-                        return Command::done(Message::NewMenu {
-                            settings: IcedNewMenuSettings {
-                                size: (100, 100),
-                                direction: MenuDirection::Up,
-                            },
-                            id,
-                        });
+                        if let Some(parent) = self.window_id(&WindowInfo::Left).copied() {
```
Decodetalkers: It is not acceptable, then it becomes every time I click the right button, the popup will always shown on the Left panel. We should not replay on a parent root, it should be found after being called

Decodetalkers (changes requested):

iced_layershell/src/actions.rs
```diff
@@ -25,19 +29,55 @@ impl From<IcedXdgWindowSettings> for NewXdgWindowSettings {
...
+    /// (in the parent surface's local coordinates) on the `parent` surface.
+    ///
+    /// Defaults are applied: anchored at the bottom-left of the anchor
+    /// rect, growing toward the top-right, with the compositor free to flip
+    /// or slide the popup on either axis to keep it on-screen. Override any of
+    /// them with the builder methods.
+    pub fn new(parent: IcedId, size: (u32, u32), anchor_rect: (i32, i32, i32, i32)) -> Self {
+        Self {
+            size,
+            parent,
```
Decodetalkers, line 68: we should remove it. parent should not be decided by the user

iced_layershell/src/multi_window.rs
```diff
@@ -810,17 +810,29 @@ where
...
-                let Some(parent_layer_shell_id) = ev.current_surface_id() else {
+                let IcedNewPopupSettings {
+                    size,
+                    parent,
+                    anchor_rect,
+                    anchor,
+                    gravity,
+                    constraint_adjustment,
+                } = settings;
+                let Some(parent_layer_id) = self.window_manager.get(parent).map(|w| w.id) else {
```
Decodetalkers: use current_surface_id instead

exwlshellev/src/events.rs
Decodetalkers: I see , I wrong

iced_layershell/src/actions.rs
```diff
@@ -25,19 +29,55 @@ impl From<IcedXdgWindowSettings> for NewXdgWindowSettings {
...
-}
+impl IcedNewPopupSettings {
+    /// Create popup settings for a popup of `size` anchored at `anchor_rect`
+    /// (in the parent surface's local coordinates) on the `parent` surface.
+    ///
+    /// Defaults are applied: anchored at the bottom-left of the anchor
+    /// rect, growing toward the top-right, with the compositor free to flip
+    /// or slide the popup on either axis to keep it on-screen. Override any of
+    /// them with the builder methods.
+    pub fn new(parent: IcedId, size: (u32, u32), anchor_rect: (i32, i32, i32, i32)) -> Self {
```
Decodetalkers, line 49: I think should we need anchor_rect at the beginning. I mean we can give it a default value, and use something like
```rust
fn anchor_react(self, rect: impl Into<AnchorReact>) -> Self {

}
```

to define it later

iced_layershell/src/multi_window.rs
```diff
@@ -829,37 +841,6 @@ where
                     Some(iced_id),
                 )));
             }
-            LayerShellCustomAction::NewMenu {
```
Decodetalkers, line 832: I want to keep it. We just want the popup to be shown at the clicked position. Please keep it

I just want a menu, so I added the popup. Do not delete it

iced_exwlshell/src/multi_window.rs
id3v1669: > we do not need to care about what the layer is

For xdg_popups with grab we do. Anchor rect can go wrong, nesting can go wrong. In general there are lot's of relationship with parent surface

> if not set

For that case I agree that good idea to have it as a fallback, or opt-in option, but not default.

Something like that should work:
```rust
let parent_layer_id = match parent {
  Some(parent) => self.window_manager.get(parent).map(|w| w.id),
  None => ev.current_surface_id(),
};
let Some(parent_layer_id) = parent_layer_id else {
```
alongside with making parent an `Option`

Fallback is enough or we need explicit option?

iced_wayland_subscriber/README.md
id3v1669: Yes, the idea was to show in example explicit parent definition.

iced_layershell/src/build_pattern/daemon.md
id3v1669: Not sure about that one, layers already have margin and it works comply different from anchor rect, that probably would only add confusion.

For better readability we can create a separate struct, but not sure how that makes much improvement

iced_layershell/src/actions.rs
id3v1669: Nested popups required parent that is set by user(at least logically).

In comments for code in multi_window we would make it optional, but there are many cases where setting parent is needed like multi-window apps where app tries to spawn a popup with exact parent regardless of where user focus is.

iced_layershell/src/multi_window/state.rs
```diff
@@ -145,10 +145,6 @@ where
         self.theme_mode
     }
 
-    pub fn mouse_position(&self) -> Option<&Point> {
```
Decodetalkers, line 148: Then can we bring back the menu option first? That one is also needed.

iced_examples/counter_multi/src/main.rs
Decodetalkers: I want to keep an example for the menu. After right clicked, it will show a popup according to the mouse position. The example was for that purpose

id3v1669:
@Decodetalkers
Done, brought back NewMenu. In counter_multi it spawns now the same way as before.

iced_exwlshell/src/multi_window.rs
```diff
@@ -865,26 +882,31 @@ where
                 let Some((_, window)) = self.window_manager.get_alias(parent_layer_shell_id) else {
                     return;
                 };
-
                 let Some(point) = window.state.mouse_position() else {
                     return;
                 };
-
                 let (x, mut y) = (point.x as i32, point.y as i32);
                 if let MenuDirection::Up = menu_setting.direction {
```
Decodetalkers, line 889: I wonder if we can just use the anchor instead of the menudirection.. Can we remove the MenuDirection now? Is it useless?

Any way, all other things are ok

id3v1669: I mean all things related to Menu can be implemented working directly with popup, so it's up to you to decide as it's mostly about style of lib.

Decodetalkers: I see..  I will accept it.

Decodetalkers (approved):

Decodetalkers:
Anyway, I will do the last things, like adjust the code styles, and etc. Thank you very much

id3v1669:
@Decodetalkers 
I've made one mistake that was fixed locally, but I forgot it to add to repo before committing.
In layereshellev/src/lib.rs and exwlshellev/src/lib.rs
From:
```rust
pub fn take_popup_grab_serial(&mut self) -> Option<u32> {
        self.button_serial.take()
    }
```
To:
```rust
pub fn take_popup_grab_serial(&mut self) -> Option<u32> {
        self.button_serial.take().or(self.enter_serial)
    }
```
Do you minde commiting that so no extra pr created?

That code is needed for chained popups that are triggered by actions like hover behave correctly.

Decodetalkers:
> @Decodetalkers 
> I've made one mistake that was fixed locally, but I forgot it to add to repo before committing.
> In layereshellev/src/lib.rs and exwlshellev/src/lib.rs
> From:
> ```rust
> pub fn take_popup_grab_serial(&mut self) -> Option<u32> {
>         self.button_serial.take()
>     }
> ```
> To:
> ```rust
> pub fn take_popup_grab_serial(&mut self) -> Option<u32> {
>         self.button_serial.take().or(self.enter_serial)
>     }
> ```
> Do you minde commiting that so no extra pr created?
> 
> That code is needed for chained popups that are triggered by actions like hover behave correctly.

Ok

Decodetalkers:
You have helped me a lot.. If you are glad, can I invite you to comaintain this project? I still not that familiar with wayland protocols. So if you come to my help, that will help me a lot!

id3v1669:
> You have helped me a lot.. If you are glad, can I invite you to comaintain this project? I still not that familiar with wayland protocols. So if you come to my help, that will help me a lot!

I'm also not that familiar with wayland protocols, I just dive into docs when I get into into issues with things I code or use. So feel free to tag on github or discord(same name and avatar) if you need help.
I'm not sure if inviting me to co-maintain is a good idea as I don't always have time to code and my contributions will be very inconsistent. Tagging me would be better option.

Decodetalkers:
> > You have helped me a lot.. If you are glad, can I invite you to comaintain this project? I still not that familiar with wayland protocols. So if you come to my help, that will help me a lot!
> 
> I'm also not that familiar with wayland protocols, I just dive into docs when I get into into issues with things I code or use. So feel free to tag on github or discord(same name and avatar) if you need help.
> I'm not sure if inviting me to co-maintain is a good idea as I don't always have time to code and my contributions will be very inconsistent. Tagging me would be better option.

Ok,still thank you very much!

PR #387, opened 2026-06-12
refactor: share keymap conversion implementation

httpsworldview:
This moves the duplicated keyboard conversion tables into iced_exdevtools::keymap.
The idea is that iced_exwlshell, iced_layershell, iced_sessionlock, should share a single implementation.
Reduces duplication, also removes unused smol_str dependencies from those crates and adds winit-core where the shared helper now exists.


Decodetalkers (approved):

Decodetalkers:
Thanks!

PR #392, opened 2026-07-06
fix: None option cannot work in layershellev and exwlshellev, and replace None with Active

id3v1669:
A way to let compositor to decide output, just pass `NULL` to `get_layer_surface`.

Refrence: [ref](https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/blob/master/unstable%2Fwlr-layer-shell-unstable-v1.xml?blame=1#L56)

### DIFF
--- a/exwlshellev/src/events.rs
+++ b/exwlshellev/src/events.rs
@@ -84,6 +84,10 @@ pub enum OutputOption {
     /// want to pass a [wl_output::WlOutput] to create a new layershell, you need to pass your
     /// connection to layershellev first
     Output(wl_output::WlOutput),
+
+    /// Let the compositor decide which output to use.
+    Active,
+
     #[default]
     None,
 }
@@ -165,7 +169,7 @@ impl Default for NewLayerShellSettings {
             size: None,
             margin: Some((0, 0, 0, 0)),
             keyboard_interactivity: KeyboardInteractivity::OnDemand,
-            output_option: OutputOption::None,
+            output_option: OutputOption::Active,
             events_transparent: false,
             namespace: None,
         }
--- a/exwlshellev/src/events.rs
+++ b/exwlshellev/src/events.rs
@@ -86,10 +86,8 @@ pub enum OutputOption {
     Output(wl_output::WlOutput),
 
     /// Let the compositor decide which output to use.
-    Active,
-
     #[default]
-    None,
+    Active,
 }
 
 /// layershell settings to create a new layershell surface
--- a/exwlshellev/src/lib.rs
+++ b/exwlshellev/src/lib.rs
@@ -1086,6 +1086,17 @@ impl<T> WindowState<T> {
     pub fn forget_last_output(&mut self) {
         self.last_wloutput.take();
     }
+
+    fn last_output(&mut self) -> Option<WlOutput> {
+        if self.last_wloutput.is_none() {
+            self.last_wloutput = self.outputs.get(self.last_unit_index).cloned();
+        }
+
+        self.last_wloutput
+            .as_ref()
+            .or_else(|| self.outputs.first())
+            .cloned()
+    }
 }
 
 /// Simple WindowState, without any data binding or info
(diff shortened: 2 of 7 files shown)


Decodetalkers:
We already have an OutputOption::None. I used to use it as active, but some days the meanings changed.  Can we replace the None with Active? It should be a mistake

id3v1669:
> We already have an OutputOption::None. I used to use it as active, but some days the meanings changed. Can we replace the None with Active? It should be a mistake

I think it's best to keep both as is, at least for now.
It's two implementations of the same thing, but one let's compositor to decide(as it should) and another does that on lib side. 
Keeping both gives time to make sure that this option actually works correctly and helps to avoid another breaking change(if we rename None to Active).

ps. tested on my side and it works, but to be safe

Decodetalkers:
> > We already have an OutputOption::None. I used to use it as active, but some days the meanings changed. Can we replace the None with Active? It should be a mistake
> 
> I think it's best to keep both as is, at least for now.
> It's two implementations of the same thing, but one let's compositor to decide(as it should) and another does that on lib side. 
> Keeping both gives time to make sure that this option actually works correctly and helps to avoid another breaking change(if we rename None to Active).

We already have many breaking changes these days... And now the None option is meaningless. Before I also did not know why None will not work.. if you care about the breaking changes, I think you can let None and Active to be the same thing, and expand that match block. I do not want to keep a meaningless options. When did it broken... I forgot why I change logic to that. None should be active. But maybe active will be a better name

id3v1669:
Something like that?

httpsworldview:
I think `LastOutput` is accidentally broken here? The new `_ => None` branch also catches `OutputOption::LastOutput`, so it behaves like `Active` instead of using the remembered output. For input panels it then hits `let Some(output) = output else { ... continue; }`, so creation is skipped.

Should we match `Active`/`None` explicitly and keep the old path for `LastOutput`?

Decodetalkers:
No, what you changed is wrong. Like @httpsworldview , you just need to expand https://github.com/waycrate/exwlshelleventloop/blob/master/layershellev/src/lib.rs#L2618-L2643 this part. Before it should only be used for LastOutput, but I mistook this part, change the behavior of None.

What I want to say is expand this part, make Active to the default, and keep active and None to the same behavior, I just thought you do not want to induce a breaking change, so I am thinking about to add a deprecate to tell user to use Active instead.

For example

```rust
                            let output = match output_type {
                                OutputOption::Output(output) => Some(output),
                                OutputOption::OutputName(name) => window_state
                                    .xdg_info_cache
                                    .iter()
                                    .find(|(_, info)| info.name == *name)
                                    .map(|(output, _)| output.clone()),
                                OutputOption::Active | Output::None => None,
                                OutputOption::LastOutput => {
                                    let pos = window_state.surface_pos();

                                    let mut output =
                                        pos.and_then(|p| window_state.units[p].wl_output.as_ref());

                                    if window_state.last_wloutput.is_none()
                                        && window_state.outputs.len() > window_state.last_unit_index
                                    {
                                        window_state.last_wloutput = Some(
                                            window_state.outputs[window_state.last_unit_index]
                                                .clone(),
                                        );
                                    }

                                    output = window_state.last_wloutput.as_ref();

                                    if output.is_none() {
                                        output = window_state.outputs.first();
                                    }

                                    output.cloned()
                                }
                            };
```

Decodetalkers:
But since we have already induced a lot of breaking changes these time, last release it should be another breaking change release, so I think it is safe to rename None to Active

httpsworldview:
@Decodetalkers does that (a4444b4) look alright?

Decodetalkers (approved):
LGTM, but before we need to change the title

Decodetalkers:
Ok, maybe we need to publish a new release soon

id3v1669:
> @Decodetalkers does that ([a4444b4](https://github.com/waycrate/exwlshelleventloop/commit/a4444b4a615e696f33aa761344a73a5da07b52ad)) look alright?

Thanks for wrapping pr up and apologize for my inattention yesterday.

Decodetalkers:
> 
> 
> 
> > @Decodetalkers does that ([a4444b4](https://github.com/waycrate/exwlshelleventloop/commit/a4444b4a615e696f33aa761344a73a5da07b52ad)) look alright?
> 
> Thanks for wrapping pr up and apologize for my inattention yesterday. 
> 

No problem! It was us need to thank you! Thank you for fix this bug for me.

PR #396, opened 2026-07-16
feat: support ext-background-effect-v1

Decodetalkers:
land support for ext-background-effect

haven't tested yet, because this protocol is not landed on sway,

### DIFF
--- a/iced_layershell_macros/src/lib.rs
+++ b/iced_layershell_macros/src/lib.rs
@@ -43,6 +43,7 @@ pub fn to_layer_message(attr: TokenStream2, input: TokenStream2) -> manyhow::Res
                 LayerChange{id: iced_layershell::reexport::IcedId, layer:iced_layershell::reexport::Layer},
                 /// Margin: top, left, bottom, right
                 MarginChange{id: iced_layershell::reexport::IcedId, margin: (i32, i32, i32, i32)},
+                BlurOptionChange{id: iced_layershell::reexport::IcedId, option: iced_layershell::reexport::BlurOption},
                 SizeChange{id: iced_layershell::reexport::IcedId, size: (u32, u32)},
                 ExclusiveZoneChange{id: iced_layershell::reexport::IcedId, zone_size: i32},
                 KeyboardInteractivityChange{id: iced_layershell::reexport::IcedId, keyboard_interactivity: iced_layershell::reexport::KeyboardInteractivity},
@@ -120,6 +121,7 @@ pub fn to_layer_message(attr: TokenStream2, input: TokenStream2) -> manyhow::Res
                             Self::NewInputPanel {settings, id } => Ok(LayerShellCustomActionWithId::new(None, LayerShellCustomAction::NewInputPanel { settings, id })),
                             Self::RemoveWindow(id) => Ok(LayerShellCustomActionWithId::new(Some(id), LayerShellCustomAction::RemoveWindow)),
                             Self::ForgetLastOutput => Ok(LayerShellCustomActionWithId::new(None, LayerShellCustomAction::ForgetLastOutput)),
+                            Self::BlurOptionChange {id, option} => Ok(LayerShellCustomActionWithId::new(Some(id), LayerShellCustomAction::BlurOptionChange(option))),
                             _ => Err(self)
                         }
                     }
@@ -142,6 +144,7 @@ pub fn to_layer_message(attr: TokenStream2, input: TokenStream2) -> manyhow::Res
                 VirtualKeyboardPressed {
                     key: u32,
                 },
+                BlurOptionChange(iced_layershell::reexport::BlurOption),
             };
             let impl_quote = quote! {
                 impl #impl_gen TryInto<iced_layershell::actions::LayerShellCustomActionWithId> for #ident #ty_gen #where_gen {
@@ -164,6 +167,8 @@ pub fn to_layer_message(attr: TokenStream2, input: TokenStream2) -> manyhow::Res
                             Self::VirtualKeyboardPressed { key } => Ok(LayerShellCustomActionWithId::new(None, LayerShellCustomAction::VirtualKeyboardPressed {
                                 key
                             })),
+
+                            Self::BlurOptionChange(option) => Ok(LayerShellCustomActionWithId::new(None, LayerShellCustomAction::BlurOptionChange(option))),
                             _ => Err(self)
                         }
                     }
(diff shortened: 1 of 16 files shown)


Decodetalkers:
@httpsworldview sorry for bother you, do you have any opinion on this pr? I won't land it for a while

layershellev/src/lib.rs
```diff
@@ -657,6 +680,34 @@ impl<T> WindowStateUnit<T> {
         }
     }
 
+    pub fn set_blur_option(&mut self, blur_option: BlurOption) {
+        self.blur_option = blur_option;
+        if let Some(effect) = &self.effect {
+            match &self.blur_option {
+                BlurOption::None => {}
```
httpsworldview: I believe this leaves the previously committed region active. The protocol requires set_blur_region(None) to remove the effect.

layershellev/src/lib.rs
```diff
@@ -657,6 +680,34 @@ impl<T> WindowStateUnit<T> {
...
+                    for BlurRegion {
+                        x,
+                        y,
+                        width,
+                        height,
+                    } in regions
+                    {
+                        region.add(*x, *y, *width, *height);
+                    }
+                    effect.set_blur_region(Some(&region));
```
httpsworldview, line 638: 690-here leaks a wl_region on every update. protocol has copy semantics, so the region should probably be destroyed right after

layershellev/src/lib.rs
```diff
@@ -657,6 +680,34 @@ impl<T> WindowStateUnit<T> {
...
+                }
+            }
+        }
+    }
```
httpsworldview, line 644: set_blur_option never commits, so an idle window may never apply a dynamic change

layershellev/src/lib.rs
```diff
@@ -3013,6 +3098,29 @@ impl<T: 'static> WindowState<T> {
...
+                                    width,
+                                    height,
+                                } in regions
+                                {
+                                    region.add(*x, *y, *width, *height);
+                                }
+                                effect.set_blur_region(Some(&region));
+                            }
+                        }
+                    }
```
httpsworldview, line 3018: R3098 commits the buffer before set_blur_region, so the initial blur remains pending

layershellev/src/lib.rs
httpsworldview, line 2273: Leaves the builder defaults, which means initial windows discard the configured blur.

httpsworldview, line 2342: Same as R2365

iced_layershell_macros/src/lib.rs
```diff
@@ -120,6 +121,7 @@ pub fn to_layer_message(attr: TokenStream2, input: TokenStream2) -> manyhow::Res
                             Self::NewInputPanel {settings, id } => Ok(LayerShellCustomActionWithId::new(None, LayerShellCustomAction::NewInputPanel { settings, id })),
                             Self::RemoveWindow(id) => Ok(LayerShellCustomActionWithId::new(Some(id), LayerShellCustomAction::RemoveWindow)),
                             Self::ForgetLastOutput => Ok(LayerShellCustomActionWithId::new(None, LayerShellCustomAction::ForgetLastOutput)),
+                            Self::BlurOptionChange {id, option} => Ok(LayerShellCustomActionWithId::new(Some(id), LayerShellCustomAction::BlurOption(option))),
```
httpsworldview:
```suggestion
                            Self::BlurOptionChange {id, option} => Ok(LayerShellCustomActionWithId::new(Some(id), LayerShellCustomAction::BlurOptionChange(option))),
```

iced_layershell_macros/src/lib.rs
```diff
@@ -142,6 +144,7 @@ pub fn to_layer_message(attr: TokenStream2, input: TokenStream2) -> manyhow::Res
                 VirtualKeyboardPressed {
                     key: u32,
                 },
+                BlurOptionChange(iced_layershell::reexport::BlurOptionChange),
```
httpsworldview:
```suggestion
                BlurOptionChange(iced_layershell::reexport::BlurOption),
```

iced_layershell_macros/src/lib.rs
```diff
@@ -164,6 +167,8 @@ pub fn to_layer_message(attr: TokenStream2, input: TokenStream2) -> manyhow::Res
                             Self::VirtualKeyboardPressed { key } => Ok(LayerShellCustomActionWithId::new(None, LayerShellCustomAction::VirtualKeyboardPressed {
                                 key
                             })),
+
+                            Self::BlurOptionChange(option) => Ok(LayerShellCustomActionWithId::new(None, LayerShellCustomAction::BlurOption(option))),
```
httpsworldview:
```suggestion
                            Self::BlurOptionChange(option) => Ok(LayerShellCustomActionWithId::new(None, LayerShellCustomAction::BlurOptionChange(option))),
```

httpsworldview (commented):
I think this is good so far, but too render-loop centric considering blur is Wayland surface state. I'm not sure what the best way to do things would look like.

Decodetalkers:
Thank you for the review.. but I do not know how to test this pr.. Have one wm implemented this?

httpsworldview:
I think Hyprland implements the protocol on main, but it hasn't reached a stable release yet. I can try to compile and test when I get home :]

I'll let you know

id3v1669:
> Thank you for the review.. but I do not know how to test this pr.. Have one wm implemented this?

Tested on latest commit of hyprland, works.
Before:
Pasted image (2)

After:
Pasted image

httpsworldview:
Beat me to it! Thank you. I also verified that it works on Niri 26.04 ([above initial blur beahvior](https://github.com/waycrate/exwlshelleventloop/pull/396#discussion_r3598389247) withstanding)

Decodetalkers:
I wonder if the Region option will work.. I am thinking about just blur an element of iced , every time it will report its postion and size, then call the function to set its blur.. I think then wm will just blur its part

id3v1669:
> I wonder if the Region option will work.. I am thinking about just blur an element of iced , every time it will report its postion and size, then call the function to set its blur.. I think then wm will just blur its part

just tested with hardcoded values for testing purposes(Works):
Pasted image (3)

Both were spawned with same values: 
```rust
blur_option: BlurOption::Region(vec![BlurRegion {
            x: 50,
            y: 24,
            width: 300,
            height: 48,
        }]),
```

Regarding pos and size, we can use `iced::advanced::widget::operate` and get containers(rectangles) to blur. That way we could blur specific iced elements. 
The problem are rounded elements. If there is a transparent bg around blured element we would need element to be a rectangle(not rounded) - no idea how to fix that yet.

Decodetalkers:
ok, I will also make this feature on exwlshell

Decodetalkers:
and in fact , exwlshell is the fully support for both layershell and session lock, is fit for shell implement, I think, so it has most the same logic with layershell and session lock implement.. emmm

Decodetalkers:
ok, merge it this time

PR #397, opened 2026-07-21
feat: track outputs via sctk OutputState & add OutputChanged event

id3v1669:
done:

- Replace manual zxdg with sctk's `OutputState` (breaking: removes `XdgInfoChanged`, `XdgInfoChangedType`, `ZxdgOutputInfo`, `get_xdgoutput_info()`)
- new `DispatchMessage::OutputChanged` sent when a surface enters another output or info changes
- new fns under `WindowState`: `outputs()`, `output_by_name()`, `get_output_info()`, `get_output_info_of()`, `WindowStateUnit::get_wloutput()`
- new `output::listen()` subscription to track on what output window is displayed

Why/How discovered issues:
wanted to have multi-surface widgets that have weird spawn timings and initially spawn via `Active`, but other surfaces need to inherit output from first widget without depending on mouse pos.

### DIFF
--- a/exwlshellev/src/events.rs
+++ b/exwlshellev/src/events.rs
@@ -46,8 +46,6 @@ pub enum ExWlShellEvent<'a, T, Message> {
     /// [ReturnData::RequestBind], then it will continue to the next request.
     /// Here only the above two [ReturnData] are acceptable.
     InitRequest,
-    /// if the info of the XdgOutput is changed, it will send the event
-    XdgInfoChanged(XdgInfoChangedType),
     /// After you return [ReturnData::RequestBind] in the [LayerShellEvent::InitRequest] stage, next
     /// event is [LayerShellEvent::BindProvide], you can use the GlobalList and QueueHandle to create
     /// new wayland objects.
@@ -205,15 +203,6 @@ pub enum ReturnData<INFO> {
     None,
 }
 
-/// this tell the what kind of information passed by [LayerShellEvent::XdgInfoChanged]
-#[derive(Debug, Clone, Copy)]
-pub enum XdgInfoChangedType {
-    Position,
-    Size,
-    Name,
-    Description,
-}
-
 /// Describes a scroll along one axis
 #[derive(Default, Debug, Clone, Copy, PartialEq)]
 pub struct AxisScroll {
@@ -339,7 +328,7 @@ pub(crate) enum DispatchMessageInner {
         scale_u32: u32,
         scale_float: f64,
     },
-    XdgInfoChanged(XdgInfoChangedType),
+    OutputChanged(Option<WlOutput>),
     Ime(Ime),
 }
 
@@ -438,6 +427,8 @@ pub enum DispatchMessage {
         scale_float: f64,
     },
     Ime(Ime),
+    /// surface entered output, or left the one it was on
+    OutputChanged(Option<WlOutput>),
     Closed,
 }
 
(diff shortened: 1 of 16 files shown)


iced_exwlshell/src/output.rs
```diff
@@ -0,0 +1,85 @@
...
+
+/// output where surface is displayed
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct OutputInfo {
+    /// output name, like `HDMI-A-1`
+    pub name: String,
+    /// human readable output description
+    pub description: String,
+    /// global output pos
+    pub logical_position: (i32, i32),
```
Decodetalkers: Use Size instead

iced_layershell/src/output.rs
```diff
@@ -0,0 +1,85 @@
...
+
+/// output where surface is displayed
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct OutputInfo {
+    /// output name, like `HDMI-A-1`
+    pub name: String,
+    /// human readable output description
+    pub description: String,
+    /// global output pos
+    pub logical_position: (i32, i32),
```
Decodetalkers: Same with above

exwlshellev/src/events.rs
```diff
@@ -339,7 +328,7 @@ pub(crate) enum DispatchMessageInner {
         scale_u32: u32,
         scale_float: f64,
     },
-    XdgInfoChanged(XdgInfoChangedType),
+    OutputChanged(Option<WlOutput>),
```
Decodetalkers, line 331: why it needs to be Option?

iced_exwlshell/src/output.rs
id3v1669: You mean for `logical_size`?

exwlshellev/src/events.rs
id3v1669: Because when we disconnect display we get `leave surface` event and new one might not land yet, so we get none for that period until `enter` event lands.

iced_exwlshell/src/output.rs
Decodetalkers: I mean use iced::Size

Decodetalkers: https://docs.rs/iced/0.14.0/iced/struct.Point.html

And position we can use this one

Decodetalkers: I just hate the too many tuple usage in layershellev, but I do not have that motivation to change them. at least I want to make the tuple usage less in iced crate

Decodetalkers:
Then we can change the logic here https://github.com/waycrate/exwlshelleventloop/blob/master/iced_examples/counter_multi/src/main.rs#L185-L199 to the listen inside the crate, instead of using another crate

id3v1669:
> Then we can change the logic here https://github.com/waycrate/exwlshelleventloop/blob/master/iced_examples/counter_multi/src/main.rs#L185-L199 to the listen inside the crate, instead of using another crate

It feels wrong.
OutputInsert looks like tracks new output appeared.
OutputChanged tracks move of already existing surfaces through outputs and info change for outputs where surfaces are.
We can't replace one by another.

Decodetalkers:
> > Then we can change the logic here https://github.com/waycrate/exwlshelleventloop/blob/master/iced_examples/counter_multi/src/main.rs#L185-L199 to the listen inside the crate, instead of using another crate
> 
> It feels wrong. OutputInsert looks like tracks new output appeared. OutputChanged tracks move of already existing surfaces through outputs and info change for outputs where surfaces are. We can't replace one by another.

I see. sure. I understand

Decodetalkers (approved):
LGTM

Decodetalkers:
merged

PR #402, opened 2026-07-25
feat: init guards for zero values

id3v1669:
PR to provide better guards for size & anchor values passed to lib. previously had to zero-guard dynamic values on client side when it should be on lib side.

### DIFF
--- a/exwlshellev/src/events.rs
+++ b/exwlshellev/src/events.rs
@@ -20,6 +20,8 @@ use wayland_client::{
     },
 };
 
+use crate::size::{LayerSize, PixelSize};
+
 use crate::{blur::BlurOption, id, xkb_keyboard::KeyEvent};
 
 use crate::keyboard::ModifiersState;
@@ -91,8 +93,8 @@ pub enum OutputOption {
 /// layershell settings to create a new layershell surface
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct NewLayerShellSettings {
-    /// the size of the layershell, optional.
-    pub size: Option<(u32, u32)>,
+    /// the size of the layershell
+    pub size: LayerSize,
     pub layer: Layer,
     pub anchor: Anchor,
     pub exclusive_zone: Option<i32>,
@@ -110,8 +112,13 @@ pub struct NewLayerShellSettings {
 /// How a popup is positioned relative to its parent surface.
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub enum PopupPlacement {
-    /// anchor rectangle in the parent surface's local coordinates (x, y, w, h)
-    Anchored((i32, i32, i32, i32)),
+    /// anchor rectangle in the parent surface's local coordinates
+    Anchored {
+        /// the top-left corner of the anchor rectangle
+        position: (i32, i32),
+        /// the extents of the anchor rectangle
+        size: PixelSize,
+    },
     /// Absolute position of the popup in the parent surface's local coordinates
     Position((i32, i32)),
 }
@@ -120,7 +127,7 @@ pub enum PopupPlacement {
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub struct NewPopUpSettings {
     /// the size of the popup
-    pub size: (u32, u32),
+    pub size: PixelSize,
     /// the id of the parent surface
     pub id: id::Id,
     /// How a popup is positioned relative to its parent surface.
@@ -140,7 +147,7 @@ pub struct NewXdgWindowSettings {
     /// The window title.
     pub title: Option<String>,
     /// The initial window size.
-    pub size: Option<(u32, u32)>,
+    pub size: Option<PixelSize>,
     /// Request client-side decorations instead of the default server-side mode.
     pub client_side_decorations: bool,
 }
@@ -148,7 +155,7 @@ pub struct NewXdgWindowSettings {
 /// input panel settings to create a new input panel surface
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct NewInputPanelSettings {
-    pub size: (u32, u32),
+    pub size: PixelSize,
     /// set the surface type as a keyboard
     pub keyboard: bool,
     /// follow the last output of the activated surface, used to create some thing like mako, who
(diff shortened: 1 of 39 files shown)


id3v1669:
@Decodetalkers can you check structs please. Any ideas of handling that better way, or proceed with what we have?

iced_examples/counter_timer/src/main.rs
```diff
@@ -75,22 +75,22 @@ fn update(counter: &mut Counter, message: Message, _time: Instant) -> Command<Me
...
-            WindowDirection::Bottom => Command::done(Message::AnchorSizeChange(
-                Anchor::Bottom | Anchor::Left | Anchor::Right,
-                (0, 400),
-            )),
-            WindowDirection::Top => Command::done(Message::AnchorSizeChange(
-                Anchor::Top | Anchor::Left | Anchor::Right,
-                (0, 400),
-            )),
+            WindowDirection::Left => Command::done(Message::LayoutChange {
+                anchor: Anchor::Left,
```
Decodetalkers, line 79: isn't the meaning changed?

Decodetalkers: I see. but when we want Anchor::Right | Anchor::Left?

id3v1669: No, in old example we stretched via anchors and here we have guard via fill fn to ensure correct anchors are applied. They still can be defined, just if we miss correct anchor it will get set instead of wl error

iced_examples/application_launcher/src/main.rs
```diff
@@ -24,7 +24,7 @@ fn main() -> Result<(), iced_layershell::Error> {
     )
     .settings(Settings {
         layer_settings: LayerShellSettings {
-            size: Some((1000, 1000)),
+            size: LayerSize::px(1000, 1000),
```
Decodetalkers, line 27: sometimes I do not want to set the size, want to let wm decide it for me. For example, I have set the Anchor::all(), this time I should not set the size. Is there a none option?

layershellev/src/lib.rs
```diff
@@ -2482,13 +2529,14 @@ impl<T: 'static> WindowState<T> {
...
                             (),
                         );
-                        layer.set_anchor(window_state.anchor);
+                        let wire_anchor = window_state.size.resolve_anchor(window_state.anchor);
+                        layer.set_anchor(wire_anchor);
                         layer.set_keyboard_interactivity(window_state.keyboard_interactivity);
-                        if let Some((init_w, init_h)) = window_state.size {
-                            layer.set_size(init_w, init_h);
-                        }
+                        let (init_w, init_h) = window_state.size.to_set();
```
Decodetalkers, line 2535: this part the meaning is changed.  we should not always set the size.

iced_examples/application_launcher/src/main.rs
id3v1669: That's what `FILL` is for. It set's zeros. I've read that: "If you pass 0 for either value, the compositor will assign it", so passing zero does that. Fill functions to exactly that on 0 value, that's why they need opposite anchors to have some space, so compositor would assign something.

layershellev/src/lib.rs
id3v1669: As in prev comment, we should send zeros to let compositor decide

Decodetalkers: I see. set it to zero is the same as not setting anything. I see.

Decodetalkers:
I think two size is good. Layershell should be treated with a different way. I think it is a good solution

id3v1669:
> I think two size is good. Layershell should be treated with a different way. I think it is a good solution

Thaks will proceed with comments and the rest of lib. Question: do you mind if comments for functions will be longer? I tried to follow style of lib and keep them one-line as majority here, but it feels like some logic can be missed then.

Decodetalkers:
> > I think two size is good. Layershell should be treated with a different way. I think it is a good solution
> 
> Thaks will proceed with comments and the rest of lib. Question: do you mind if comments for functions will be longer? I tried to follow style of lib and keep them one-line as majority here, but it feels like some logic can be missed then.

I do not mind, just do it freely..

exwlshellev/src/events.rs
```diff
@@ -160,10 +167,10 @@ pub struct NewInputPanelSettings {
 impl Default for NewLayerShellSettings {
     fn default() -> Self {
         NewLayerShellSettings {
-            anchor: Anchor::Bottom | Anchor::Left | Anchor::Right,
+            anchor: Anchor::Bottom | Anchor::Left | Anchor::Right | Anchor::Top,
```
Decodetalkers: you can use Anchor::all() instead

Decodetalkers:
For this change, we need to write an document to tell users to sync their code. I think

id3v1669:
> For this change, we need to write an document to tell users to sync their code. I think

Like a migration guide or make it part of changelog?
Anyway just tell me where to put it and I'll write one.

exwlshellev/src/events.rs
id3v1669: done

Decodetalkers (approved):
LGTM

Decodetalkers:
I think we should add a migration guide for this change, like what bevy does. Anyway, this pr is fine. Thank you very much

id3v1669:
> I think we should add a migration guide for this change, like what bevy does. Anyway, this pr is fine. Thank you very much

plz drop the refrense, not familiar with what you are talking about.

Also another question not related to this pr:
I want to impl subscription to workspaces via ext-workspace-v1 protocol and not sure if that repo is correct place for that.
I'll use it alongside iced_layershell, but it doesn't depend on it, so that's why thinking whether create a separate repo for it or you don't mind having it here too.

Decodetalkers:
> > I think we should add a migration guide for this change, like what bevy does. Anyway, this pr is fine. Thank you very much
> 
> plz drop the refrense, not familiar with what you are talking about.
> 
> Also another question not related to this pr:
> I want to impl subscription to workspaces via ext-workspace-v1 protocol and not sure if that repo is correct place for that.
> I'll use it alongside iced_layershell, but it doesn't depend on it, so that's why thinking whether create a separate repo for it or you don't mind having it here too.

You can add it to iced-wayland-subscriber. That is the place for extra wayland events

PR #404, opened 2026-07-28
feat: xdg_popup::reposition

id3v1669:
Add popup reposition that allows to reposition and resize popup without destroying and recreating it.
Also fix leak I previously PRed, I forgot to destroy positioner after popup creation.

### DIFF
--- a/exwlshellev/src/events.rs
+++ b/exwlshellev/src/events.rs
@@ -141,6 +141,22 @@ pub struct NewPopUpSettings {
     /// Serial of the input event
     pub grab_serial: Option<u32>,
 }
+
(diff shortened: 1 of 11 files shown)


Decodetalkers:
Maybe we need an example for this feature..

id3v1669:
> Maybe we need an example for this feature..

I added example to counter_multi

Decodetalkers:
> > Maybe we need an example for this feature..
> 
> I added example to counter_multi

Aaaa.. I see

Decodetalkers (approved):

Decodetalkers:
Thanks

PR #407, opened 2026-08-02
feat: ext-workspace-v1 as feature & rework wayland subscriber

id3v1669:
Major changes:

- `ext-workspace-v1` client behind a non-default `workspace` feature with example `workspace_bar`.
- sctk `OutputState` replaces hand-written `zxdg_output_v1`. Outputs are released on teardown.
- `ShellInfo` trait and its macro-injected variant removed. replaced by the opt-in `on_new_shell` hook plus `Settings::shell_broadcast`
- `shell.rs`: replaces two identical per-crate copies.
- `info.rs`: `OutputInfo` re-exported from sctk
- `worker.rs`: The previous poll did `roundtrip()` then `wake_by_ref(); Poll::Pending`.
Now wayland init and dispatch run on a dedicated named thread with `blocking_dispatch`.

A few notes:

- move of ownership to the Arc-shared `WindowWrapper` was because new `Drop` destroyed the `wl_surface` while wgpu still held its raw pointer

Emit all extra events via broadcast, never in update:
A broadcast message arrives at minimum one frame later, and the window is built and presented in that gap. A session-lock surface must render lock UI on its first frame. Shipped both: a synchronous `on_new_shell` hook for first-frame correctness, the async broadcast for everyone else.

Shared broadcast emitting `NewDisplay`:
`output::listen` already reports displays from a real wayland connection with live proxies and full sctk information.

"Shared" broadcast:
Built as an explicit `channel()` pair rather than a process-global.

Most questionable chage is use of MAX_RESTARTS, as statet in comment:
shouldn't be an issue. I would keep it that way and make changes only if issue created.

ps. Also working on migration guide, will add it to this pr, please tell me where to put `MIGRATION.md` file when I'm done with it.
pps. some change were hotfixes of problems I ran into, while testing, so might seem unrelated to main theme of pr

STILL WIP, NO REVIEW YET REQUIRED


iced_exwlshell/src/build_pattern/daemon.rs
```diff
@@ -664,19 +662,24 @@ impl<P: Program> Daemon<P> {
...
-        let program = {
+        let (program, on_new_shell) = {
             iced_debug::init(iced_debug::Metadata {
                 name: P::name(),
                 theme: None,
                 can_time_travel: cfg!(feature = "time-travel"),
             });
 
-            super::attach(self.raw)
+            let hook = on_new_shell.map(|f| {
```
Decodetalkers, line 678: I would prefer use register, and receive event that place, without add a new api..

Decodetalkers:
What I meant before is that broadcast the events that layershellev , sessionlockev themselves can already know to the broadcast in the origin "output", that means you can make that subscription broadcast more events

And something like workspace which is not related to the layershellev or sessionlockev, they should be put in iced_wayland_subscription. 

And I would not prefer to induce a new api named new_shell for it, I would prefer subscription, and handle the events in update. This is my opinions

Decodetalkers:
https://github.com/waycrate/exwlshelleventloop/blob/master/iced_exwlshell/src/output.rs

Since layershellev already knows a lot of events, so that events should not be broadcasted by another wayland register. We should broadcast it directly. 

And the changes also seem a bit too much.. Before I think just broadcast the needed events in the origin `output.rs` (maybe it needs to be renamed to wayland.rs), and address the newShell events in update, it will be finished, the changes won't be that much, And we can remove nearly all the logic in original iced_wayland_subscription, and just add the extra protocols which does not related to the render itself..

id3v1669:
Busy week and will be able to properly respond and continue working on PR only on Wednesday. 
Briefly looking through your responses I can reply fast on that:

> And I would not prefer to induce a new api named new_shell for it, I would prefer subscription, and handle the events in update. This is my opinions

Regarding new_shell, through a subscription the message arrives a frame after the surface is first presented, so the first frame renders the fallback branch. It's ok for now since lock screens draw the same thing on every output, but as soon as we broadcast per-surface info and apps start rendering per-monitor it would be a problem. So I'd keep the hook.

Decodetalkers:
Ok, I think you are right.. if we make them in the same Subscription, we will receive a lot of update events.. let me think about that

Decodetalkers:
Ok, your design will be better I think. Thanks

id3v1669:
@Decodetalkers 
I'm not sure if we keep `output::listen` or not. On one hand it can be used by regular iced app to get output info, but we don't impl full xdg toplevel stuff. On another hand we have minimal toplevel. We either should port missing stuff and drop `output::listen` or keep it. 
There were way too many changes already, so I would rather keep it and decide on it later.
ps. I still didn't fully update readme files and still working on docs. 
pps. I'll re-review for mistakes that I might've missed this week, but codewise it's mostly done.
About file names, we can decide later, but I would prefer to keep `output.rs` as we might add more protocols and just to avoid confusion.

id3v1669:
@Decodetalkers  I think it's ready for review, readme files are updated. You didn't respond about where to put migration guide, so I've sent it to you in discord dms.

Decodetalkers (approved):
I have understood the new design and the problems. so let us ship it

Decodetalkers:
I think it is fine. Merged.

PR #411, opened 2026-08-14
fix: create compositor synchronously & opt-in keep_compositor_alive

id3v1669:
- init compositor synchronously as there is no reason to over-complicate that. Also follows iced winit and lowers the delay on first surface open for daemon mode
- option not to kill compositor on last surface kill. Helps with "first" surface start on daemon. It has a cost in terms of gpu/ram, so made it opt-in instead of opt-out.

Also continuing conversation of previous PRs:
I would keep both listen and local top-level impl
Reason for that is not to complicate life for cross-platform devs who use winit of iced and keep an option to expand toplevel in the future for linux-only devs.

### DIFF
--- a/iced_exwlshell/src/multi_window.rs
+++ b/iced_exwlshell/src/multi_window.rs
@@ -26,7 +26,6 @@ use exwlshellev::{
 };
 #[cfg(all(feature = "linux-theme-detection", target_os = "linux"))]
 use futures::StreamExt;
-use futures::{FutureExt, future::LocalBoxFuture};
 #[cfg(not(all(feature = "linux-theme-detection", target_os = "linux")))]
 use iced_core::theme::Mode;
 use iced_core::{
@@ -48,7 +47,6 @@ use std::{
     mem,
     os::fd::AsFd,
     sync::Arc,
-    task::Poll,
     time::Duration,
 };
 use window_manager::Window;
@@ -91,7 +89,6 @@ where
     P::Message: 'static + TryInto<ExwlShellCustomActionWithId, Error = P::Message>,
 {
     use exwlshellev::calloop::channel::channel;
-    use futures::task;
     let (message_sender, message_receiver) = channel::<Action<P::Message>>();
 
     let boot_span = iced_debug::boot();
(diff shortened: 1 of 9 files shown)


iced_layershell/src/multi_window.rs
```diff
@@ -355,15 +342,13 @@ where
...
-        display: DisplayWrapper,
-    ) -> Self {
+    /// Create compositor synchronously. This is a one-time init that must finish
+    /// before the first frame can render. Copies iced_winit logic.
+    fn create_compositor(&mut self, window: Arc<WindowWrapper>, display: DisplayWrapper) {
         let shell = Shell::new(self.proxy.clone());
-        let mut new_compositor = C::new(self.compositor_settings, display, window.clone(), shell)
-            .await
-            .expect("Cannot create compositer");
+        let compositor_future = C::new(self.compositor_settings, display, window.clone(), shell);
```
Decodetalkers, line 349: yes , it is the same in iced, confirmed

iced_exwlshell/src/multi_window.rs
```diff
@@ -730,7 +712,8 @@ where
                 status: iced_core::event::Status::Ignored,
             });
         // if now there is no windows now, then break the compositor, and unlink the clipboard
-        if self.window_manager.is_empty() {
+        // unless opted-in to keep the compositor alive
+        if self.window_manager.is_empty() && !self.keep_compositor_alive {
```
Decodetalkers, line 716: Why we need to keep the compositer alive.. Before if there is no surface exist, the origin window is also dead, the new window won't react to the mouse event and keyboard events, so this logic is here, in order to create a new activated one

Decodetalkers:
With iced 0.14, it needs a display to create a compositor, so the problem is gone, we do not need to remove compositor any more I think. We can just delete that logic directly

iced_exwlshell/src/multi_window.rs
id3v1669: because compositor creation is on of those things that creates a visible delay. mouse and keyboard reactivity live in layershellev and they should survive. I've been using workaround to bypass those issues by creating untouchable 1px presistent surface on daemon mode, just finaly decided to fix the root of the problem.

Decodetalkers: now you can delete this logic directly, we do not need it anymore, I have confirmed this logic

id3v1669:
> just delete that logic directly

Maybe just set it do default and make an opt-out then? it still might be usefull for some people who don't care about delay, but care about resources. But that's up to you, I only need decreased delay option anyway

Decodetalkers:
> > just delete that logic directly
> 
> Maybe just set it do default and make an opt-out then? it still might be usefull for some people who don't care about delay, but care about resources. But that's up to you, I only need decreased delay option anyway

it just a logic to keep the new surface can work before iced 0.14, now the new compositor will always work, so I think it is fine.. it is up to you

Decodetalkers:
I prefer the delete the remove_compositor logic.. or just make the option default to true?

Decodetalkers (approved):
LGTM

PR #412, opened 2026-08-15
feat(iced): make proc macros optional but enabled by default

httpsworldview:
## Summary

Make the proc macros within `iced_layershell` and sister crates optional. Consumers that do not use the message macros can now avoid their proc-macro dependency chains, that's all.

Existing users with `default_features = false` who use macros must now enable macros explicitly, which might need to be added to the migration docs.

### DIFF
--- a/iced_exwlshell/Cargo.toml
+++ b/iced_exwlshell/Cargo.toml
@@ -11,8 +11,9 @@ description = "The binding of extra shells on wayland for iced"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-default = ["linux-theme-detection"]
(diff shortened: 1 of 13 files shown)


Decodetalkers:
reasonable, but it will be a bit troublesome to implement manually...

What problem have you meet then you want to make it optional? I just want to listen about it

Decodetalkers (approved):
LGTM

httpsworldview:
My motivation was just dependency hygiene. The macro crates were unconditional dependencies, so cargo would pull in `darling`, `manyhow`, `syn`, `quote`, etc. regardless of whether or not the consumer actually used the macros. Still left them enabled by default to avoid changing ergonomics for normal users

PR #413, opened 2026-08-18
refactor: deprecate the iced_sessionlock and etc

Decodetalkers:
remove the logic to iced_exwlshell, that one contains all of it. made a application function in iced_exwlshell for easy migration


iced_exwlshell/src/build_pattern/sessionlock.rs
id3v1669, line 1: can't we have here something like:
```rust
use super::super::daemon::{
        BootFn, NameSpace, UpdateFn, with_executor, with_scale_factor, with_style,
        with_subscription, with_theme,
    };
```
Instead of keeping local with_executor and other stuff?

id3v1669 (changes requested):

iced_exwlshell/src/build_pattern/sessionlock.rs
Decodetalkers: done, but with_theme used the id, so it cannot be reusable

iced_exwlshell/src/build_pattern.rs
```diff
@@ -12,4 +12,6 @@ pub use daemon::daemon;
 
 pub use daemon::Daemon;
```
id3v1669, line 14: add #[doc = include_str!("./build_pattern/sessionlock.md")]

Decodetalkers: done

id3v1669 (approved):

id3v1669:
Thanks

Decodetalkers:
Oh.. I should make you the maintainer

PR #415, opened 2026-08-19
refactor: merge layershell to iced_exwlshell and etc

Decodetalkers:
since exwlshell has all logic related to layershell, and can be started without sessionlock, and if only used with sessionlock, it will not panic because of layershell, so it is safe to merge the two into it.

This pr move the application and timer logic into iced_exwlshell, move the to_layershell_message to iced_exwlshell_macros crate. When using `to_layershell_message`, the detail of sessionlock will be hidden, so that the logic of sessionlock won't be triggered


Decodetalkers:
@httpsworldview after considering a lot, I think we should also deprecate the iced_layershell and move the logic to iced_exwlshell, the logic is almost the same, but just let user to use another crate

httpsworldview:
Yes, I think so too. I was working on a similar refactor locally and I think this is the right direction.

httpsworldview:
I am at work currently, but I will try to review some changes. Once I am home in a few hours I can help more. Thank you!

id3v1669:
> I am at work currently

Same, will review at ~20:00 AWST

iced_layershell/src/ime_preedit.rs
```diff
@@ -1,148 +0,0 @@
...
-            Some(selection) => {
-                vec![
-                    text::Span::new(&preedit.content[..selection.start]),
-                    text::Span::new(if selection.start == selection.end {
-                        "\u{200A}"
-                    } else {
-                        &preedit.content[selection.start..selection.end]
-                    }),
-                    text::Span::new(&preedit.content[selection.end..]),
-                ]
```
httpsworldview: iced_exwlshell still uses the window background as selected preedit text color, which was fixed here in PR #381 but not applied uniformly. both ime_preedit.rs and window_manager.rs changes should be ported.

httpsworldview (commented):
## regressions

- `exwlshellev/src/lib.rs:2433-2441` dispatches seat events before
  assigning `self.text_input_manager`;
  `exwlshellev/src/seat.rs:151-157` can then consume the keyboard
  capability while the manager is `None`, never creating
  text-input-v3.
- `exwlshellev/src/lib.rs:1355-1360` updates `self.text_input`, which
  is initialized to `None` and never assigned.
- `iced_exwlshell/src/ime_preedit.rs` and
  `iced_exwlshell/src/multi_window/window_manager.rs` still use the
  window background as selected preedit text color; PR #381 fixed both
  only in the deleted `iced_layershell`.
- `exwlshellev/src/lib.rs:3048` stores `window_state.anchor` and
  `window_state.size` for a dynamic layer surface instead of the
  requested anchor and size.
- `exwlshellev/src/lib.rs:2783-2850` creates AllScreens surfaces for
  hotplugged outputs without their configured blur option or
  background-effect object.
- `iced_exwlshell/src/actions.rs:83-90` changes the layer-shell popup
  default from `PopupGravity::TopRight` to `BottomRight`, moving
  popups whose callers relied on the former default.

## API and docs

- `iced_exwlshell::layershell::application` publicly returns
  `SingleApplication`, but the type is only privately imported and
  external users cannot name it.
- `docs/MIGRATION-19.1->20.0.md:200` references the nonexistent
  `to_layershell_message` macro instead of `to_layer_message`.
- `docs/MIGRATION-19.1->20.0.md:196` says `iced_exwlshell` was removed
  when the removed crate is `iced_layershell`.
- `iced_wayland_subscriber/README.md:13-17` still imports and returns
  errors from the removed `iced_layershell` crate.
- `iced_exwlshell/src/build_pattern/layershell.rs:1` imports
  debug-only `attach` unconditionally, producing an unused-import
  warning in standalone default builds.

## Existing issues

Mostly minor.

- `exwlshellev/src/seat.rs:179-181` destroys a seat's text input when
  any capability is removed, so losing pointer or touch destroys IME
  support without a keyboard transition to recreate it.
- `exwlshellev` commits new lock surfaces before assigning their
  session-lock role, permitting the protocol's `already_constructed`
  error.
- The runtime ignores the session lock's `locked` and `finished`
  events, then calls `unlock_and_destroy` without knowing whether the
  protocol permits it.
- Replacing the session-lock builder's settings with
  `Settings::default()` changes `StartMode::Background` and can
  trigger its start-mode assertion.

Decodetalkers:
> ## regressions
> 
>     * `exwlshellev/src/lib.rs:2433-2441` dispatches seat events before
>       assigning `self.text_input_manager`;
>       `exwlshellev/src/seat.rs:151-157` can then consume the keyboard
>       capability while the manager is `None`, never creating
>       text-input-v3.
> 
>     * `exwlshellev/src/lib.rs:1355-1360` updates `self.text_input`, which
>       is initialized to `None` and never assigned.
> 
>     * `iced_exwlshell/src/ime_preedit.rs` and
>       `iced_exwlshell/src/multi_window/window_manager.rs` still use the
>       window background as selected preedit text color; PR #381 fixed both
>       only in the deleted `iced_layershell`.
> 
>     * `exwlshellev/src/lib.rs:3048` stores `window_state.anchor` and
>       `window_state.size` for a dynamic layer surface instead of the
>       requested anchor and size.
> 
>     * `exwlshellev/src/lib.rs:2783-2850` creates AllScreens surfaces for
>       hotplugged outputs without their configured blur option or
>       background-effect object.
> 
>     * `iced_exwlshell/src/actions.rs:83-90` changes the layer-shell popup
>       default from `PopupGravity::TopRight` to `BottomRight`, moving
>       popups whose callers relied on the former default.
> 
> 
> ## API and docs
> 
>     * `iced_exwlshell::layershell::application` publicly returns
>       `SingleApplication`, but the type is only privately imported and
>       external users cannot name it.
> 
>     * `docs/MIGRATION-19.1->20.0.md:200` references the nonexistent
>       `to_layershell_message` macro instead of `to_layer_message`.
> 
>     * `docs/MIGRATION-19.1->20.0.md:196` says `iced_exwlshell` was removed
>       when the removed crate is `iced_layershell`.
> 
>     * `iced_wayland_subscriber/README.md:13-17` still imports and returns
>       errors from the removed `iced_layershell` crate.
> 
>     * `iced_exwlshell/src/build_pattern/layershell.rs:1` imports
>       debug-only `attach` unconditionally, producing an unused-import
>       warning in standalone default builds.
> 
> 
> ## Existing issues
> 
> Mostly minor.
> 
>     * `exwlshellev/src/seat.rs:179-181` destroys a seat's text input when
>       any capability is removed, so losing pointer or touch destroys IME
>       support without a keyboard transition to recreate it.
> 
>     * `exwlshellev` commits new lock surfaces before assigning their
>       session-lock role, permitting the protocol's `already_constructed`
>       error.
> 
>     * The runtime ignores the session lock's `locked` and `finished`
>       events, then calls `unlock_and_destroy` without knowing whether the
>       protocol permits it.
> 
>     * Replacing the session-lock builder's settings with
>       `Settings::default()` changes `StartMode::Background` and can
>       trigger its start-mode assertion.

You helped me a lot, but I still want to suggest you do not believe in what ai said.. 

There are lots of problems in this reply, at least

```

iced_exwlshell::layershell::application publicly returns
SingleApplication, but the type is only privately imported and
external users cannot name it.

```

This is a technology called sealed, just expose the usage of the trait, but not trait itself. it is quite used in iced and many other places. I think ai hasn't learnt about this technology, but these days rust community are designing a new seal way, but I still prefer this.

About the part of seat, that time the first dispatch also never happens, that won't influence anything.

What I want to hear about is if this change will help you or still some problems you want to comment

Decodetalkers:
About the settings.. I am thinking about make a new settings for the sessionlock part, in order not to let user edit the layershell part. but since this pr is for layershell, so it is not included

httpsworldview:
Yes, I apologize. It is late, and I took a cheap shortcut, you are right. I don't see very many problems, this does seem largely beneficial and would make working with the library easier. Sorry again

Decodetalkers:
> Yes, I apologize. It is late, and I took a cheap shortcut, you are right. I don't see very many problems, this does seem largely beneficial and would make working with the library easier. Sorry again

no apologization, take it eazy! it is not bad to receive help from tools, but I think even get help from other place, it should become your things first, and on github timeline, I want to talk with developers, that makes my day!

httpsworldview:
> it should become your things first, and on github timeline, I want to talk with developers, that makes my day!

I agree very much so. i'll strive to be better, thank you for your guidance, you are a big inspiration!

Decodetalkers:
ok, what ai said was right, the text-input-v3 was broken, because I run the dispatch before it. Thank you very much.. I made a mistake.

I fogot the input-manager was created during seat. and that part was copied from sessionlock, so that place there were two blocking_dispatch

Thank you very much

Decodetalkers:
ok. should not always believe in ai.. about text-input it was right.. but before declare it was the lock, it should commit first, at least on sway maybe because I did not handle the lock event.. but at least it works, if no report, maybe it will be better not to change it

Decodetalkers:
Ok, I should look into it

Decodetalkers:
ok, maybe chatgpt is clever than me. I think I have misunderstood the protocol, maybe nowadays the sessionlock is works on the bug of the sway.. ok, Thanks for your check @httpsworldview . I need to read the document more carefully next time

Decodetalkers:
it is kind wired.. according to the document, I should commit it after I received the I bind it to the sessionlock, but if I commit that place, it disable the rules, so I need to wait for the ExxtSessionLock surface. 

But with xdgwmbase and layershell, I must to, then it won't enter the configure place. But their documents are all the same..

Decodetalkers:
So I should understand it as when the lock is ready, it will auto send configure event for me, so I do not need to commit anything 

it works till today because I have set a empty commit, so it caused no problems..

Decodetalkers:
Ok, still need a commit before request a lock surface.. or the surface on new display won't show.. have add comment for it.

Decodetalkers:
Maybe inconvenient for us to separate the iced::Settings and our settings, we do not only have the LayerShell settings and also WithConnection, and etc.. That won't let the count of params down, because we need to pass more params into it..

About sessionlock part, I just use the assert with information, because it can be fixed quickly. If users set it manually, they will know the problem immediately. 

I think it is good to keep the part of Settings as it is now.

exwlshellev/src/lib.rs
```diff
@@ -2849,6 +2875,73 @@ impl<T: 'static> WindowState<T> {
...
+                            }
+                            window_state.handle_event(
+                                &mut *event_handler,
+                                ExWlShellEvent::RequestMessages(&DispatchMessage::Locked),
+                                None,
+                            );
+                        }
+                        (_, DispatchMessageInner::LockFinished) => {
+                            if let Some(lock) = lock.take() {
+                                lock.unlock_and_destroy();
```
id3v1669: I see we unlock and destroy, but we souldn't do that if we are not locked, just destroy if not locked. We should either make a check if we are locked or probably would be better to make a struct that holds a state or something like that. We only check if lock request was made, so it would be Some

exwlshellev/src/lib.rs
```diff
@@ -2849,6 +2875,73 @@ impl<T: 'static> WindowState<T> {
                                 .build(),
                             );
                         }
+                        (_, DispatchMessageInner::Locked) => {
+                            let l_lock = lock.as_ref().unwrap();
```
id3v1669: Are you sure we always have it here?
different sofware but similar scenario might happen. At work we have usb keys and daemon for them that auto-unlocks the software. If we send lock, but it get's auto-unlocked externaly we might not have it here. Not sure if that's the case here, but got to mention

exwlshellev/src/lib.rs
Decodetalkers: If lock is some, that means it is locked. So I think this logic is right..

exwlshellev/src/lib.rs
Decodetalkers: When the lock event is sent, we should already have a lock, if not, it will be a problem.

This event is when lock happened, then the message come back, so I think here there will always be one

exwlshellev/src/lib.rs
id3v1669: I thought that means that we got response from compositor and not the fact that it's locked

exwlshellev/src/lib.rs
id3v1669: i mean race against external unlocker, not against software that uses lib

Decodetalkers: Ok, let me check if it is some first. Then do the lock

exwlshellev/src/lib.rs
```diff
@@ -2892,47 +2985,6 @@ impl<T: 'static> WindowState<T> {
                                     continue;
                                 };
                                 let l_lock = lock_manager.lock(&qh, ());
-                                let wl_outputs = window_state.outputs.clone();
```
id3v1669, line 2895: Also wasn't that here as guard to prevent from compositors backup appearing before our layershell arrives?

Decodetalkers: I have read this document, https://wayland.app/protocols/ext-session-lock-v1#ext_session_lock_v1:request:get_lock_surface , so I was worrying if my implement is right or wrong

exwlshellev/src/lib.rs
Decodetalkers: but if lock is there, shouldn't we unlock it I think

Decodetalkers: This might be sent because there is already another ext_session_lock_v1 object held by a client, or the compositor has decided to deny the request to lock the session for some other reason. This might also be sent because the compositor implements some alternative, secure way to authenticate and unlock the session.

exwlshellev/src/lib.rs
id3v1669: Yes and it says there `These won't be displayed by the compositor unless the lock is successful and the locked event is sent.`
I think that implies that we need to create them correctly before even trying to lock

exwlshellev/src/lib.rs
Decodetalkers: anyway, if we have request a lock, and the compositor decline it for some reason, we have to unlock and destroy it. And it is also the reason that the surface need to be created after lock event come. because lock can be declined

exwlshellev/src/lib.rs
id3v1669: Maybe the location is semi-wrong, but it still needs to be done

Decodetalkers: ok, 
```
The client is expected to create lock surfaces for all outputs currently present and any new outputs as they are advertised. These won't be displayed by the compositor unless the lock is successful and the locked event is sent.
```

you are right.. I cared about useless things

Decodetalkers: ok, I will put them back

exwlshellev/src/lib.rs
id3v1669: Can't we get a response that we are not successfully locked? In that case we should just destroy and not try to unlock.

id3v1669: I hate github, you replied 4 min ago and it updated only after I replied

Decodetalkers: it does have error, https://wayland.app/protocols/ext-session-lock-v1#ext_session_lock_v1:enum:error, but seems cannot get after request.. do you have better solution?

id3v1669: those protocol errors aren't catchable anyway, compositor just kills the client
maybe turn lock into enum and just make metods for it to avoid sending wrong requests? that would prevent such calls and all will be handled in one place...
something like:
enum LockLifecycle {
    Unlocked,
    Pending { lock: ExtSessionLockV1, teardown_requested: Option<Teardown> },
    Locked  { lock: ExtSessionLockV1 },
}

id3v1669: ps. the teardown is to track what is legal to send

id3v1669: I have an idea on how to handle that, will pr to that branch in a few hours

id3v1669 (approved):
Thank you, no more comments from me

Decodetalkers:
Thank you @id3v1669  and @httpsworldview , let me merge it

PR #416, opened 2026-08-19
Enum

id3v1669:
as per conversation in #415 
most changes are fmt, so easier to view through vscode git diff

### DIFF
--- a/iced_exwlshell/src/build_pattern/layershell.rs
+++ b/iced_exwlshell/src/build_pattern/layershell.rs
@@ -398,13 +398,14 @@ mod pattern {
                 ..iced_graphics::Settings::default()
             };
             use exwlshellev::StartMode;
-            assert!(
-                !matches!(
-                    settings.layer_settings.start_mode,
-                    StartMode::AllScreens | StartMode::Background
-                ),
-                "when using application of layershell, you should not set the mode to AllScreens and Background, please give it a display or use the Active"
-            );
+            if matches!(
+                settings.layer_settings.start_mode,
+                StartMode::AllScreens | StartMode::Background
+            ) {
+                return Err(crate::Error::InvalidSettings(
+                    "when using application of layershell, you should not set the mode to AllScreens and Background, please give it a display or use the Active",
+                ));
+            }
             crate::multi_window::run(
                 program,
                 &self.namespace,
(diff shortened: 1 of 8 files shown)


iced_exwlshell/src/build_pattern/layershell.rs
```diff
@@ -398,13 +398,14 @@ mod pattern {
...
             };
             use exwlshellev::StartMode;
-            assert!(
-                !matches!(
-                    settings.layer_settings.start_mode,
-                    StartMode::AllScreens | StartMode::Background
-                ),
-                "when using application of layershell, you should not set the mode to AllScreens and Background, please give it a display or use the Active"
-            );
+            if matches!(
```
Decodetalkers, line 401: I prefer panic, we should not let the application alive, it should let user know in order to fix it immediately. They use this mode, so they know they have set a wrong mode

iced_exwlshell/src/build_pattern/sessionlock.rs
```diff
@@ -415,8 +409,21 @@ mod pattern {
             )
         }
 
+        /// Sets the [`Settings`] of the [`SingleApplication`].
+        ///
+        /// A session lock always runs in [`StartMode::Background`], it's
+        /// forced instead of panic.
         pub fn settings(self, settings: Settings) -> Self {
-            Self { settings, ..self }
+            Self {
```
Decodetalkers, line 417: I do not think it is the right way. They should know they do wrong

iced_exwlshell/src/build_pattern/layershell.rs
id3v1669: I work on app where I declare most things on the level of config file, so there will be text var parsed to mode. That's either extra check that users like me would have to figure out for dynamic vars or check on lib side that throws meaningful error

iced_exwlshell/src/build_pattern/sessionlock.rs
id3v1669: > They should know they do wrong

Why? Isn't lib meant to make life easier and protect from solutions that break stuff?

Decodetalkers: But sometimes we need break, and I think the panic message is clear enough. And I agree with the layershell part, but I this part is unconfigable

Decodetalkers: ok, let it pass. it is also reasonable

iced_exwlshell/src/build_pattern/layershell.rs
Decodetalkers: I see

PR #417, opened 2026-08-21
chore: text-input lifetime adjustment

Decodetalkers:
text-input should always alive when a seat is created, and destroy it when seat id died

We cannot suggested text-input will always need a keyboard, because we can have other input way like virtual keyboard

### DIFF
--- a/exwlshellev/src/seat.rs
+++ b/exwlshellev/src/seat.rs
@@ -110,6 +110,10 @@ impl Drop for SeatStorage {
         {
             pointer.release();
         }
+        // destroy text_input during drop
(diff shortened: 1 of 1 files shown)


Decodetalkers:
oh.. rust has published a new version

id3v1669 (approved):
lgtm

id3v1669:
> oh.. rust has published a new version

guess I have outdated package as didn't trigger on that locally

Decodetalkers:
> > oh.. rust has published a new version
> 
> guess I have outdated package as didn't trigger on that locally

ok, now it is fixed! Thank you very much!

PR #418, opened 2026-08-21
fix: track foucs from compositor

id3v1669:
Fixes focus issues on Hyprland, please test on sway and smithhay-based like niri.

if fix works fine and no extra guards/manual handling needed I will fix virtual keyboard example(there are issues there with printing symbols)

what and how:
just tust wl_keyboard::enter for the focus
even on touch event compositor should shift focus and trigger wl_keyboard event, so we don't realy need manual tracking

### DIFF
--- a/exwlshellev/src/seat.rs
+++ b/exwlshellev/src/seat.rs
@@ -31,13 +31,6 @@ impl<T> WindowState<T> {
             .cloned()
             .collect()
     }
-    pub fn get_keyboard_state_iter_mut(
-        &mut self,
-    ) -> impl Iterator<Item = &mut xkb_keyboard::KeyboardState> {
-        self.seats
-            .values_mut()
-            .flat_map(|seat| &mut seat.keyboard_state)
-    }
     pub fn get_keyboard_state_by_id(
         &mut self,
         id: ObjectId,
@@ -219,7 +212,6 @@ impl<T> Dispatch<wl_keyboard::WlKeyboard, ()> for WindowState<T> {
     ) {
         use crate::keyboard::*;
         use xkb_keyboard::ElementState;
-        let surface_id = state.current_surface_id();
 
         match event {
             wl_keyboard::Event::Keymap { format, fd, size } => match format {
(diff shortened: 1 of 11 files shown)


Decodetalkers:
So this pr separated the focus of pointer and keyboard, and pass the focus event as the iced focus event change. I used tried to use keyboard focus, but that time the auto focus of application launcher did not work, or did that time I mixed two focuses? Let me take a try tomorrow

id3v1669:
> So this pr separated the focus of pointer and keyboard, and pass the focus event as the iced focus event change. I used tried to use keyboard focus, but that time the auto focus of application launcher did not work, or did that time I mixed two focuses? Let me take a try tomorrow

I'm still working on correction of pointer focus as it's still buggy. Current changes just make wl_keyboard work correctly and hopefuly do not make other things worse

id3v1669:
Last commit should fix #334 
At least works for text field and cursor preserves it's shape based on pointer pos

id3v1669:
I'll fix virtual keyboard example separately as it requires lib adjustment and that's for a separate pr.
Ready for review

exwlshellev/src/seat.rs
```diff
@@ -246,47 +238,42 @@ impl<T> Dispatch<wl_keyboard::WlKeyboard, ()> for WindowState<T> {
...
-                    .find(|seat_storage| {
-                        seat_storage
-                            .keyboard_state
-                            .as_ref()
-                            .is_some_and(|state| state.keyboard == *wl_keyboard)
-                    })
-                    .map(|storage| storage.keyboard_state.as_mut().unwrap())
-                else {
+            wl_keyboard::Event::Leave { surface, .. } => {
+                state.keyboard_focus = None;
```
httpsworldview, line 262: Should keyboard focus be stored per keyboard/seat instead of globally? I think if two keyboards focus a surface, one leaving would clear the global focus and cause the other's subsequent inputs to be routed without the surface's ID. Not sure how realistic that situation is or if it's really a problem...

id3v1669: Wlroots has support for multi-seat. Unlike xorg milti-seat it's not that well supported and been in development for years, but that's where per-seat wl_keyboard would be usefull.
  ps. still waiting for that functionality to work properly with multi-gpu configuration(unrelated to issue, but blocking multi-seat usage for me)

Decodetalkers:
Looks good for me, and didn't broke anything. 

But what problem have you meet with the original implement? I did think the original implement may not be a good one

id3v1669:
> Looks good for me, and didn't broke anything.
> 
> But what problem have you meet with the original implement? I did think the original implement may not be a good one

you can check the example that tracks focus pre my commits.
With single window I could never catch focus, just unfocus event arrived,
When second window spawned, only then focus event started working,
With 2 windows when I click outside of window(any window that doesn't belong to us) focus became buggy again intil 2-3 focus/unfocus events.

Decodetalkers (approved):
LGTM, it is really a problem. let's ship it

PR #419, opened 2026-08-23
fix: memory leak if create too many child windows

Decodetalkers:
resolve: https://github.com/waycrate/exwlshelleventloop/issues/346

the shells used to create new surfaces should not be created for too many times and without being destroyed.


httpsworldview (commented):
This looks good to me, but closed surfaces don't clean up pending frame callbacks, because the WlCallback returned by wl_surface.frame() is discarded.

should we handle that here too??

Decodetalkers:
> This looks good to me, but closed surfaces don't clean up pending frame callbacks, because the WlCallback returned by wl_surface.frame() is discarded.
> 
> should we handle that here too??

You are right

Decodetalkers:
> This looks good to me, but closed surfaces don't clean up pending frame callbacks, because the WlCallback returned by wl_surface.frame() is discarded.
> 
> should we handle that here too??

done. I have included your part in this pr

httpsworldview (approved):
Awesome, looks good to me

PR #420, opened 2026-08-24
feat: add an unique action to skip refresh

### LINKED ISSUE #403: Add support for on-demand redrawing
Currently, a `Message` update triggers a repaint request for all surfaces. For instance, if Surface A has a 60 FPS animation and Surface B is a static panel, Surface A updates its animation progress via `Message` every frame. Because the message path calls `request_refresh_all()`, Surface B will also continuously receive refresh requests, resulting in a massive amount of unnecessary repaints.

Therefore, I suggest adding a new API that can request a repaint based on a specific `window_id`. I would be more than happy to help implement this if needed.

Decodetalkers:
resolve: https://github.com/waycrate/exwlshelleventloop/issues/403

sometimes after we subscrbe an action for example timer, it will refresh every second.

This action allow you to control such action, when you think it is not the time to refresh the surface, send this action


Decodetalkers:
@krteke what do you think about this solution?

krteke:
First of all, I am so sorry for missing your previous [comment](https://github.com/waycrate/exwlshelleventloop/issues/403#issuecomment-5126695178) on my issue😭😭😭. It got buried in the CI notifications, and I have been quite busy lately and didn't see it until now.

I took a look at the code. `NoRefresh` is very useful for messages that don't need to repaint any surface, but it only allows a choice between "refresh all surfaces" and "refresh no surface". However, this doesn't actually solve the problem I mentioned earlier.

Also, there is another issue, the condition here [seems to be reversed](https://github.com/waycrate/exwlshelleventloop/pull/420/changes#diff-509d933aa5e1ead218f49dbad9f79ed0f13e7f134aede19155468a2eae2e7355R1269):
```rust
if !matches!(action, ExwlShellCustomAction::NoRefresh)
```

Now I believe these two features can actually complement each other perfectly. `NoRefresh` is great for messages that need zero repainting, while the `redraw_scope` approach can be used to target specific windows.

If you think this combined approach makes sense, I'd be happy to submit a PR for it. I'm a bit swamped with other things at the moment, so it might take a few days before I can open the PR.

Decodetalkers:
> First of all, I am so sorry for missing your previous [comment](https://github.com/waycrate/exwlshelleventloop/issues/403#issuecomment-5126695178) on my issue😭😭😭. It got buried in the CI notifications, and I have been quite busy lately and didn't see it until now.
> 
> I took a look at the code. `NoRefresh` is very useful for messages that don't need to repaint any surface, but it only allows a choice between "refresh all surfaces" and "refresh no surface". However, this doesn't actually solve the problem I mentioned earlier.
> 
> Also, there is another issue, the condition here [seems to be reversed](https://github.com/waycrate/exwlshelleventloop/pull/420/changes#diff-509d933aa5e1ead218f49dbad9f79ed0f13e7f134aede19155468a2eae2e7355R1269):
> 
> ```rust
> if !matches!(action, ExwlShellCustomAction::NoRefresh)
> ```
> 
> Now I believe these two features can actually complement each other perfectly. `NoRefresh` is great for messages that need zero repainting, while the `redraw_scope` approach can be used to target specific windows.
> 
> If you think this combined approach makes sense, I'd be happy to submit a PR for it. I'm a bit swamped with other things at the moment, so it might take a few days before I can open the PR.

Because we can get the result through result , and we also can make Messages to our Custom command, can we add the policy as an extension for Message? I do not that want to add one new interface.. You are right, we need to handle for situations, all, one, many, none, can we add it to the Commands instead add an new interface?

I will be busy this week, I will be glad that if you can continue the job. 

What do you think about this? Will happy to receive your reply

krteke:
I've just analyzed this a bit more carefully. If we follow the current approach, implementing things like `RefreshOne`, `RefreshMany`, etc., entirely via Commands may work fine.

However, `NoRefresh` doesn't quite work this way because it cannot accurately track which `Message` it is supposed to prevent from triggering a redraw. If the intention is to use `NoRefresh` to cancel a redraw triggered by a previous `Message`, we would have to do something like `NoRefresh(Message)`. But doing so would introduce generic parameters to `ExwlShellCustomAction`, which would make the codebase look quite clunky and messy.

I can think of two possible alternatives to solve this:

Option 1:

- Regular `Messages` never trigger an automatic refresh.
- Every update must explicitly append a `RedrawAll`, `RedrawOne`, or `RedrawMany` command.
- If no redraw command is appended, it is simply equivalent to None (no refresh).
- Drawback: This would be a breaking change that alters the current default refresh behavior.

Option 2:

Remove `NoRefresh` entirely, and exclusively use `Message::RedrawAll/One/Many`.

What are your thoughts on these approaches?

Decodetalkers:
I see, it was my mistake, I used to thought the update was the application.update, but in fact, it is not.. so ok, I think your approach is right, I will close this pr, because it is wrong. Maybe we should induce a new api

Decodetalkers:
Ok, this pr is wrong.. Let us induce a new api for it. @krteke I think what you have done is right. Thanks for pointing out my mistake. I made a mistake

PR #421, opened 2026-08-27
feat: add configurable redraw policy to iced_exwlshell

### LINKED ISSUE #403: Add support for on-demand redrawing

krteke:
Resolves #403

Adds a configurable redraw policy for selecting all, one, or no target windows based on each message.
The existing redraw-all behavior remains the default.

### DIFF
--- a/iced_exwlshell/src/redraw.rs
+++ b/iced_exwlshell/src/redraw.rs
@@ -1,0 +1,96 @@
+use iced_core::window::Id;
+use std::fmt::Debug;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Scope {
+    None,
+    Window(Id),
+    All,
+}
+
+pub(crate) enum Targets {
+    None,
+    Window(Id),
+    All,
+    Windows(Vec<Id>),
+}
+
+type RedrawScopeFn<Message> = dyn Fn(&Message) -> Scope;
+
+pub(crate) struct Policy<Message> {
+    scope: Option<Box<RedrawScopeFn<Message>>>,
+}
+
+impl<Message> Policy<Message> {
+    pub(crate) fn new(scope: impl Fn(&Message) -> Scope + 'static) -> Self {
+        Self {
+            scope: Some(Box::new(scope)),
+        }
+    }
+
+    #[cfg(any(all(feature = "debug", not(target_arch = "wasm32")), test))]
+    pub(crate) fn for_wrapped_messages<Wrapped>(
+        self,
+        inner_message: impl for<'a> Fn(&'a Wrapped) -> Option<&'a Message> + 'static,
+    ) -> Policy<Wrapped>
+    where
+        Message: 'static,
+    {
+        let Some(scope) = self.scope else {
+            return Policy::default();
+        };
+
+        Policy::new(move |message| inner_message(message).map_or(Scope::All, &scope))
+    }
+
+    pub(crate) fn targets(&self, messages: &[Message]) -> Targets {
+        let Some(scope) = self.scope.as_ref() else {
+            return Targets::All;
+        };
+
+        let mut first_window = None;
+        let mut messages = messages.iter();
+        let mut windows = None;
+
+        for message in messages.by_ref() {
+            match scope(message) {
+                Scope::All => return Targets::All,
+                Scope::None => {}
+                Scope::Window(id) => match first_window {
+                    None => first_window = Some(id),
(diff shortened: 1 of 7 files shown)


iced_exwlshell/src/redraw.rs
```diff
@@ -0,0 +1,96 @@
...
+                    }
+                },
+            }
+        }
+
+        if let Some(mut windows) = windows {
+            for message in messages {
+                match scope(message) {
+                    Scope::Window(id) if !windows.contains(&id) => windows.push(id),
+                    Scope::All => return Targets::All,
```
Decodetalkers, line 74: I think this is already returned before, so we can use unreachable!() for this part

iced_exwlshell/src/redraw.rs
```diff
@@ -0,0 +1,96 @@
...
+                    Some(current) if current == id => {}
+                    Some(current) => {
+                        windows = Some(vec![current, id]);
+                        break;
+                    }
+                },
+            }
+        }
+
+        if let Some(mut windows) = windows {
```
Decodetalkers, line 70: why did it do twice? Can you explain it for me? thanks

iced_exwlshell/src/redraw.rs
```diff
@@ -0,0 +1,96 @@
...
+
+        for message in messages.by_ref() {
+            match scope(message) {
+                Scope::All => return Targets::All,
+                Scope::None => {}
+                Scope::Window(id) => match first_window {
+                    None => first_window = Some(id),
+                    Some(current) if current == id => {}
+                    Some(current) => {
+                        windows = Some(vec![current, id]);
```
Decodetalkers, line 63: should we wait all messages being addressed, then collect the window id?

iced_exwlshell/src/redraw.rs
krteke: This is necessary here and cannot be changed to `unreachable!()`.

For example: `win1, win2, all`.
When processing `win2`, the previous iteration sets `windows = Some(vec![win1, win2])`.
The next iteration then continues processing the remaining `all`.
Changing this to `unreachable!()` would cause a valid message batch to panic.

iced_exwlshell/src/redraw.rs
krteke: What I intended here is an optimization specifically for scenarios where a batch of messages only targets a single window for refresh ([msg1 -> win1, msg2 -> win1]). By doing this, we can avoid any unnecessary heap allocations for a `Vec<Id>`.

The overall logic of that block works like this:

1. It first tries to iterate through each message.
2. If it finds that all `Scopes` point to the same window, or they are all `Scope::None`, then `windows` remains `None`. It will ultimately just return `Window(id)` or `Targets::None`.
3. However, if it encounters a different window `Id` during this iteration, it breaks out of the loop early.
4. It then uses a new `for` loop to process the rest of the messages, and finally returns a deduplicated `Windows(Vec<Id>)`.

iced_exwlshell/src/redraw.rs
krteke: > should we wait all messages being addressed, then collect the window id?

No, this part is not for collection, because

```rust
None => first_window = Some(id),
Some(current) if current == id => {}
```

would affect the collection process.

The actual collection happens in the `for` loop below.

Decodetalkers: Why not remove Window only leave it with Target::Windows(Vec<Id>) even there is only one window? and if it is empty then return None? Then we do not need the second loop

krteke: Because I want to avoid heap allocation for `Vec<Id>` in the one window case. There appear to be two `for` loops here, but in practice `&[Message]` is only traversed once.

Decodetalkers:
Ok, I understood. Can you also add an example for it, for example, you can add it in counter_multi. This will help users know how to use it. Thank you very much. After merging this pr, I will publish a new version for it!

Decodetalkers (approved):

Decodetalkers:
Thank you, merged

PR #425, opened 2026-09-19
Add xdg toplevel controls and compositor state tracking, separate standard cursor shapes from theme names

id3v1669:
Changes:
- Better xdg toplevel support with correct state tracking and window controls
- Allow xdg toplevel windows to parent popups
- Improve input serial handling
- Add typed cursor requests and support custom theme cursors
- Fix cursor mappings and fallback handling, and reuse the cursor surface

ps. Changelog is kinda prepped, will commit it when code in general is approved


Decodetalkers (approved):
Yes, before I just make xdg toplevel as a normal dialog, so thank you for finishing this part

PR #426, opened 2026-09-20
refactor: apply a new design for exwlshellev

Decodetalkers:
this will let the Message type gone from the ExWlShellEvent, and let user can register event multi times, will make it easy for us to implement something like a11y


Decodetalkers:
No...example broken?

Decodetalkers:
I see, that example make layer all event_transparent

Decodetalkers:
Ok, done, I think it can help one day when we want to implement the a11y

Decodetalkers:
I make this change is that I think one day we need to implement a11y, but now the api is very difficult to implement it, because it binds a Message in ExWlshellEvent to pass the specific kind of message. So I take the design of winit as reference , use trait instead.

Decodetalkers:
do you have some suggestions on the design and api names? if none, I would want to merge it.. The orignal api is really a miss, so I want to make a refactor for a very long time

exwlshellev/src/lib.rs
```diff
@@ -2563,490 +2610,218 @@ impl<T: 'static> Dispatch<XdgWmBase, ()> for WindowState<T> {
...
+    fn on_init(
+        &mut self,
+        _event: ExWlShellInitEvent<T>,
+        _state: &mut WindowState<T>,
+    ) -> InitRequest {
+        InitRequest::None
+    }
 
-        let qh = event_queue.handle();
+    fn request_buffer(
```
httpsworldview, line 2628: Should we pass the target window's Id here? the old RequestBuffer received some(unit_id).

Decodetalkers: You are right

httpsworldview:
Other than that everything looks good to me.

httpsworldview (approved):

Decodetalkers:
Thanks! hope one day we can have support for a11y

id3v1669:
> Thanks! hope one day we can have support for a11y

As I can see register callbacks cannot touch WindowState, is that alright?

Decodetalkers:
> > Thanks! hope one day we can have support for a11y
> 
> As I can see register callbacks cannot touch WindowState, is that alright?

The current usage is fine, because we just forword the events , yes.. I will add a commit for it, you are right, sometimes we need it

Decodetalkers:
> > Thanks! hope one day we can have support for a11y
> 
> As I can see register callbacks cannot touch WindowState, is that alright?

Thank you, fixed.. I have merged it too fast, sorry

